PR #1, opened 2019-07-16
EVA-1586 First version of the pipeline

mistyskye:


jmmut (approved):
I don't have experience using these tools but looking at the intermediate files everything looks reasonable, but I would like to know why remapping the reference reads and adding the alt allele at the end is better than remapping the alternate reads.

README.md
```diff
@@ -1,2 +1,89 @@
 # variant-remapping
-The pipeline for remapping VCF variants between two arbitrary FASTA assemblies.
+## Description
+The pipeline for remapping VCF variants between two arbitrary FASTA assemblies. No chain file required.
```
cyenyxe:
```suggestion
Pipeline for remapping VCF variants between two arbitrary assemblies in FASTA format. No chain file is required.
```

README.md
```diff
@@ -1,2 +1,89 @@
 # variant-remapping
-The pipeline for remapping VCF variants between two arbitrary FASTA assemblies.
+## Description
```
cyenyxe: For consistency with other README files.
```suggestion
```

README.md
```diff
@@ -1,2 +1,89 @@
...
+- reverse_strand.py, included (reverse strand allele correction, uses bamnostic)
+- write_header.py, included (necessary for VCF header generation)
+- vcf2bed
+- bowtie2
+- samtools
+- bedtools
+- bcftools
+
+## Input
+- Old genome assembly file (fasta format): the genome you have variants for.
```
cyenyxe:
```suggestion
- Old genome assembly file (FASTA format): the genome you have variants for.
```

README.md
```diff
@@ -1,2 +1,89 @@
...
+- New genome assembly file (fasta format): the genome you want to remap the variants to.
```
cyenyxe:
```suggestion
- New genome assembly file (FASTA format): the genome you want to remap the variants to.
```

README.md
```diff
@@ -1,2 +1,89 @@
...
+- Variants file (vcf format): contains the list of variants you want to remap.
```
cyenyxe:
```suggestion
- Variants file (VCF format): contains the list of variants you want to remap.
```

README.md
```diff
@@ -1,2 +1,89 @@
...
+1070
+Percentage of remapped variants:
+99.300%
+
+real	0m1.363s
+user	0m1.091s
+sys	0m0.171s
+```
+
+## Task list
```
cyenyxe: Shall we create tickets in the backlog? In terms of tracking and planning, it is preferable over a list in a plain text file :)

README.md
```diff
@@ -1,2 +1,89 @@
...
-The pipeline for remapping VCF variants between two arbitrary FASTA assemblies.
+## Description
+The pipeline for remapping VCF variants between two arbitrary FASTA assemblies. No chain file required.
+
+**Method**: creates reads from the flanking sequences of each variant, then maps them to the new assembly using bowtie2.
+
+Currently, it only supports SNPs, and not indels.
+
+Prerequisites:
+- reverse_strand.py, included (reverse strand allele correction, uses bamnostic)
```
cyenyxe: Does this mean the alleles are converted to the forward strand? Could it be written down that way, for the not-so-biologically savvy (most people in the team 😂)?

remapping_commands.sh
```diff
@@ -0,0 +1,189 @@
...
+#convert vcf to bed:
+vcf2bed < snps_only.vcf > variants.bed
+#the actual position of the variant is the second coordinate
+#rm snps_only.vcf
+
+#generate the flanking sequence intervals
+bedtools flank -i variants.bed -g "$oldgenome".chrom.sizes -b 50 > flanking.bed
+rm variants.bed "$oldgenome".chrom.sizes
+
+#fix rows: put both flanking sequences on 1 row = 1 read and add 1 to start position so the read is 100 bases:
```
cyenyxe: Could you please also document briefly what it looked like before, to understand better the reason for these transformations?

remapping_commands.sh
```diff
@@ -0,0 +1,189 @@
...
+awk 'NR == 1 {prev = $2} NR>1 && NR%2==0 {print $1"\t"prev+1"\t"$3"\t"$4"\t"$5"\t"$6"\t"$7"\t"$8"\t"$9} NR%2 == 1 { prev = $2}' flanking.bed > flanking.corrected.bed
+rm flanking.bed
+
+#get the fasta sequences for these intervals
+bedtools getfasta -fi $oldgenome -bed flanking.corrected.bed -fo variants_reads.fa
+
+#replace the colon separators with "|"
+sed 's/:/|/' variants_reads.fa > variants_reads2.fa
```
cyenyxe: `sed` support in-place replacements with the `-i` argument.

remapping_commands.sh
```diff
@@ -0,0 +1,189 @@
...
+#filter SNPs only
+bcftools filter -i 'TYPE="snp"' $vcffile > snps_only.vcf
+
+#store header
+awk '$1 ~ /^#/ {print $0}' snps_only.vcf > vcf_header.txt
+
+#convert vcf to bed:
+vcf2bed < snps_only.vcf > variants.bed
+#the actual position of the variant is the second coordinate
+#rm snps_only.vcf
```
cyenyxe: Please delete commands that are not planned to be used. There are a few more in step 3.

reverse_strand.py
```diff
@@ -0,0 +1,35 @@
...
+outfile=open(args.outfile, 'w')
+
+#reverses the allele for variants that got mapped onto the reverse strand of the new genome, and prints everything into correct columns
+
+for read in bamfile:
+	name=read.read_name
+	info=name.split("|")
+	#mapped onto reverse strand:
+	if read.flag & 16: #decoding bitwise flag with &; may need to be changed back to ==
+		if info[2]=='A':
```
cyenyxe: My suggestion is to assign the reverse nucleotide in the if-else, and take the long `write` call out of them to avoid repetition.

write_header.py
```diff
@@ -0,0 +1,40 @@
+#! /usr/bin/python3
+import argparse
+from argparse import RawTextHelpFormatter
+
+description="Adds the first part of a VCF header up until the end of the ##INFOs, \nthen adds the newly generated ##configs and ##reference, and finally add the rest of the header."
```
cyenyxe: A header can be sorted any way, which means there could be useful information after an INFO entry.

reverse_strand.py
```diff
@@ -0,0 +1,35 @@
...
+
+parser = argparse.ArgumentParser(description = description, formatter_class=RawTextHelpFormatter)
+parser.add_argument("-i", "--bam", help="bam file containing remapped variants ")
+parser.add_argument("-o", "--outfile", help="name of new file")
+args = parser.parse_args()
+
+bamfile=bs.AlignmentFile(args.bam, 'rb')
+outfile=open(args.outfile, 'w')
+
+#reverses the allele for variants that got mapped onto the reverse strand of the new genome, and prints everything into correct columns
```
cyenyxe: This loop seems to complement the bases, but does it also reverse the sequence? Or is it not needed?

remapping_commands.sh
```diff
@@ -0,0 +1,189 @@
...
+#add the reference assembly
+echo "##reference="$newgenomeaccession >> contigs.txt
+
+#copy everything that isn't ##contig or ##reference from the old header to a temp
+awk '($1 !~ /^##contig/ && $1 !~ /^##reference/) {print $0}' vcf_header.txt > temp_header.txt
+rm vcf_header.txt
+
+#find the end of the INFO tags and print the contigs (using custom python script)
+chmod a+x write_header.py
+./write_header.py -i temp_header.txt -c contigs.txt -o final_header.txt
```
cyenyxe: If `vcf_header.txt` was created using only lines starting with `##`, then this Python script might not even be necessary because getting the final header would be just a matter of concatenating a bunch of other files.

cyenyxe (changes requested):

remapping_commands.sh
mistyskye: I commented the rm commands for testing purposes, but I forgot to uncomment them! I'll fix them.

reverse_strand.py
mistyskye: Good suggestion, added.

reverse_strand.py
mistyskye: By sequence do you mean the (new) reference genome? Or the read that got mapped?

write_header.py
mistyskye: I wasn't sure about the order of VCF headers, so if they're un-ordered does this mean I can just delete lines starting with ##contig from the old header, regenerate them using the new contig names, and paste them after the old header? In that case, then I agree that this Python script is unnecessary.

README.md
mistyskye: Yes that would be best :) Should I make them?

remapping_commands.sh
mistyskye: Thank you, I changed the command.

remapping_commands.sh
mistyskye: Of course, I have now added an explanation directly into the remapping_commands.sh file.

remapping_commands.sh
mistyskye: See [other comment.](https://github.com/EBIvariation/variant-remapping/pull/1/files#r308158748)

README.md
mistyskye: Yes! I've added an explanation in the README file.

remapping_commands.sh
```diff
@@ -0,0 +1,181 @@
...
+# -find a way to extract new assembly accession so it's not required as input? or just use new assembly file name?
+# -decide what to do in the case of new REF allele being the same as the variant allele (was A > G on old assembly, now is G > G on new assembly)
+
+#Completed:
+# -filter only SNPs (not indels) using bcftools filter
+# -copy filter etc columns from original vcf
+# -copy vcf header to the new vcf file, but regenerate contigs from the list of contigs in new assembly, and change reference tag
+# -take into account reverse strand
+# -create a command line version where you input file names
+# -% of remapped variants calculation
```
tskir: Given that we have a separate Markdown file with documentation, which already includes most if not all of these comments, they can be safely removed from here

remapping_commands.sh
```diff
@@ -0,0 +1,181 @@
...
+		g) oldgenome="${OPTARG}";;
+		n) newgenome="${OPTARG}" ;;
+		a) newgenomeaccession="${OPTARG}" ;;
+		v) vcffile="${OPTARG}" ;;
+		o) outfile="${OPTARG}" ;;
+	esac
+done
+
+
+#1) Flanking sequence generation
```
tskir: Since you have quite prominent `echo` statements which already define the steps, comments such as this one can be removed

remapping_commands.sh
```diff
@@ -0,0 +1,181 @@
...
+echo "-----------------------1) Flanking sequence generation-----------------------"
+#generate chrom sizes:
+samtools faidx $oldgenome 
+cut -f1,2 "$oldgenome".fai > "$oldgenome".chrom.sizes
```
tskir: In this example, `$oldgenome` is double-quoted in the second line but not in the first one. For simplicity and reliability, I suggest double-quoting all variable expansions in all contexts throughout the script

remapping_commands.sh
```diff
@@ -0,0 +1,181 @@
...
+
+#filter SNPs only
+bcftools filter -i 'TYPE="snp"' $vcffile > snps_only.vcf
+
+#store header
+awk '$1 ~ /^#/ {print $0}' snps_only.vcf > vcf_header.txt
```
tskir: FYI this can be achieved more simply by `grep '^#'` :) However, the problem with the awk/grep approach is that they will scan the entire file, which can be millions of line long. It's better to use `bcftools view --header-only` to extract the header

tskir: (It will only scan up until the `#CHROM POS ...` line and then quit)

remapping_commands.sh
```diff
@@ -0,0 +1,181 @@
...
+
+#convert vcf to bed:
+vcf2bed < snps_only.vcf > variants.bed
+#the actual position of the variant is the second coordinate
+rm snps_only.vcf
```
tskir: Given that we create *lots* of temporary files, I think it's better to store them in a temporary directory instead of just putting them in the current working directory. You can use `mktemp -d -t` to create such a directory

remapping_commands.sh
```diff
@@ -0,0 +1,181 @@
...
+
+#generate the flanking sequence intervals
+bedtools flank -i variants.bed -g "$oldgenome".chrom.sizes -b 50 > flanking.bed
```
tskir: Theoretically, even though unlikely, it is possible for a variant to be so close to the end of the chromosome that it's impossible to get a 50 bp flanking sequence for it. Could you please investigate what would happen in such a case when trying to run this command?

_This suggestion is low-priority_

remapping_commands.sh
```diff
@@ -0,0 +1,181 @@
...
+#this outputs a flanking.bed file: for 1 variant we have the following lines:
+#[chr] [-50 pos] [variant pos-1] [rsID] [qual] [REF allele] [ALT allele] [filter] [info]
+#[chr] [variant pos] [+50 pos] [rsID] [qual] [REF allele] [ALT allele] [filter] [info]
+
+#we want 1 line per variant, with 1 interval to create the read:
+#[chr] [-50 pos] [+50 pos] [rsID] [qual] [REF allele] [ALT allele] [filter] [info]
+
+#fix rows: put both flanking sequences on 1 row so each read is 101 bases: [50 bases]-[variant]-[50 bases]
+awk 'NR == 1 {prev = $2} NR>1 && NR%2==0 {print $1"\t"prev"\t"$3"\t"$4"\t"$5"\t"$6"\t"$7"\t"$8"\t"$9} NR%2 == 1 { prev = $2}' flanking.bed > flanking.corrected.bed
+rm flanking.bed
```
tskir: You can use `bedtools slop` instead of `bedtools flank` to generate a two-way flanking sequence in one go

remapping_commands.sh
```diff
@@ -0,0 +1,181 @@
...
+
+#get the fasta sequences for these intervals
+bedtools getfasta -fi $oldgenome -bed flanking.corrected.bed -fo variants_reads.fa
+
+#replace the colon separators with "|"
+sed -i 's/:/|/' variants_reads.fa
```
tskir: I'm a bit lost here, why is this step necessary? Is there some specific format of the sequence names?

remapping_commands.sh
```diff
@@ -0,0 +1,181 @@
...
+#store the other vcf columns
+awk '{print $5, $8, $9}' flanking.corrected.bed > qual_filt_info.txt
+
+#paste the names, variant bases, then fasta sequences into a new file
+paste <(grep '^>' variants_reads.fa) variant_bases.txt rsIDs.txt qual_filt_info.txt <(grep -v '^>' variants_reads.fa) > temp.txt
+rm flanking.corrected.bed variant_bases.txt rsIDs.txt qual_filt_info.txt  variants_reads.fa
+
+#reformat the fasta ID
+sed 's/\t/|/g; s/ /|/g; s/\(.*\)|/\1 /' temp.txt | tr ' ' '\n' > variant_reads.out.fa
+rm temp.txt
```
tskir: Could you please elaborate what's happening here? It would help to provide examples of input/output lines as done earlier in the script

tskir (changes requested):
The review is not yet complete; only covers documentation and Part 1 of the `remapping_commands.sh`

remapping_commands.sh
mistyskye: Bedtools flank requires a file with the length of each chromosome/contig (.chrom.sizes), this is used exactly for this purpose: 
[From the bedtools website:](https://bedtools.readthedocs.io/en/latest/content/tools/flank.html)
> Note that flank will restrict the created flanking intervals to the size of the chromosome (i.e. no start < 0 and no end > chromosome size).

So the read would be shorter, but no errors would occur.

remapping_commands.sh
mistyskye: I'm storing this information for later on in the script when I split the name by "|" to extract the relevant information (ALT allele, QUAL, FILT, INFO). I do this because the INFO column can contain ":", which means I wouldn't be able to split by ":", so I chose to use "|". Should I add this as an explanation in the script?

remapping_commands.sh
```diff
@@ -0,0 +1,181 @@
...
+
+
+
+
+
+
+#2) Mapping with bowtie2
```
tskir: I feel that 6 empty lines are a bit excessive :) A double empty line would suffice to separate the larger code blocks in my opinion

remapping_commands.sh
```diff
@@ -0,0 +1,181 @@
...
+echo "---------------------------2) Mapping with bowtie2---------------------------"
+#"index" is the prefix of the output filenames, not a directory
+echo "#### Building index: may take a while ####"
+bowtie2-build $newgenome index
+echo "#### Mapping results: ####"
+bowtie2 -f -x index variant_reads.out.fa | samtools view -bS - > reads_aligned.bam
```
tskir: 1) Could you please provide an explanation of the `bowtie21 flags used?
2) Although we have observed that Bowtie only stores one alignment per “read”, it's better to specify this behaviour explicitly (using flags)

remapping_commands.sh
```diff
@@ -0,0 +1,181 @@
...
+#3) Data extraction
+echo '------------------------------3) Data extraction-----------------------------'
+#extract chromosome, rsID, position on new genome, original variant allele, qual, filter and info fields
+samtools sort reads_aligned.bam -o reads_aligned.sorted.bam
+samtools index reads_aligned.sorted.bam
+./reverse_strand.py -i reads_aligned.sorted.bam -o variants_remapped.vcf
+rm reads_aligned.bam reads_aligned.sorted.bam reads_aligned.sorted.bam.bai
+
+#add interval for variant position in bed format (required by getfasta)
+awk '{print $1"\t"$2-1"\t"$2"\t"$3"\t"$4"\t"$5"\t"$6"\t"$7}' variants_remapped.vcf > variants_remapped.bed
```
tskir, line 130: Again this is a bit obscure, could you please provide an input-output example here?

remapping_commands.sh
```diff
@@ -0,0 +1,181 @@
...
+
+#getfasta to get the REF genome alleles and then extract the genome alleles without the positions
+bedtools getfasta -tab -fi $newgenome -bed variants_remapped.bed | awk '{print $2}'> genome_alleles.txt
+rm variants_remapped.bed
+#make all alleles caps (sometimes there are lower-case bases)
+tr a-z A-Z < genome_alleles.txt > genome_alleles.fixed.txt
+
+#paste all the information into the correct columns
+awk '{print $1"\t"$2"\t"$3}' variants_remapped.vcf > temp_first_3_columns.txt
+awk '{print $4"\t"$5"\t"$6"\t"$7}' variants_remapped.vcf > temp_last_4_columns.txt
```
tskir: When you only need to extract some columns, without doing any more complex transformations, it's easier to use `cut -f1-3` or `cut -f4-7`, respectively

remapping_commands.sh
```diff
@@ -0,0 +1,181 @@
...
+paste temp_first_3_columns.txt genome_alleles.fixed.txt temp_last_4_columns.txt > var_pre_final.vcf
+rm temp_first_3_columns.txt temp_last_4_columns.txt genome_alleles.txt variants_remapped.vcf genome_alleles.fixed.txt
+
+
+#header:
+#create list of contigs/chromosomes to be added to the header
```
tskir: I suggest you look into `bcftools reheader` to update the header information. Among other things, it can automatically update the list of contigs based on fai index. This would also the problem in current implementation where VCF header will only include contigs which have at least one variant on them (in reality it should include **all** contigs present in the assembly)

remapping_commands.sh
```diff
@@ -0,0 +1,181 @@
...
+rm temp_header.txt contigs.txt
+
+#add header to the vcf file:
+cat final_header.txt | cat - var_pre_final.vcf > temp && mv temp $outfile
+rm final_header.txt var_pre_final.vcf
+
+
+echo "-----------------------------------Results-----------------------------------"
+echo "Total number of input variants:"
+grep "^[^#]" $vcffile | wc -l
```
tskir: At least some of these statistics can be computed by `bcftools stats` instead
https://samtools.github.io/bcftools/bcftools.html#stats

reverse_strand.py
```diff
@@ -0,0 +1,36 @@
+#! /usr/bin/python3
+import argparse
+from argparse import RawTextHelpFormatter
+import bamnostic as bs
+
+description="Reverses the allele for variants that got mapped onto the reverse strand of the new genome, and prints everything to a new file\nAlso filters out the non-mapped"
```
tskir: Could you please wrap the line length throughout the code to the maximum of 120 columns?

write_header.py
tskir: Meta-information (`##`) lines can in theory go in any order, except for the very first `##fileformat` line. In the VCF specification, sections describing different meta-information lines are arranged in some order, but it never says that lines themselves must be arranged: https://samtools.github.io/hts-specs/VCFv4.2.pdf

I suggest you use `bcftools reheader` for replacing contig names

reverse_strand.py
tskir: I guess @cyenyxe meant the read. Because we're only interested in SNPs for now, we only need to flip the strand for one specific nucleotide, so actually reverse-complementing the read is unnecessary for now

reverse_strand.py
```diff
@@ -0,0 +1,36 @@
...
+		elif info[2]=='T':
+			rev_nucl="A"
+		elif info[2]=='C':
+			rev_nucl="G"
+		elif info[2]=='G':
+			rev_nucl="C"
+		outfile.write("%s\t%s\t%s\t%s\t%s\t%s\t%s\n" % (read.reference_name,read.pos+51,info[3],rev_nucl,info[4],info[5],info[6]))
+	#only mapped onto direct strand:
+	elif read.flag==0:
+		outfile.write("%s\t%s\t%s\t%s\t%s\t%s\t%s\n" % (read.reference_name,read.pos+51,info[3],info[2],info[4],info[5],info[6]))
```
tskir: This can be shortened by using https://biopython.org/wiki/Seq#complement-and-reverse-complement instead of the if-elif loop, and also only having one `outfile.write()` call. E. g.:

```python
nucl = info[2]
if read.flag & 16:
    nucl = Seq(nucl, generic_dna).complement()
outfile.write(...)
```

tskir (changes requested):
The second part of the review

README.md
tskir: Since at this point we don't have a definitive long-term roadmap for the pipeline, I think it's too soon to start creating the tickets. The only thing that we'll surely have to implement (during @mistyskye's internship or after it) is support for indels. Other points are more planning points than tickets; so I suggest just moving them to the shared Google document so far

reverse_strand.py
tskir: @mistyskye Although this discussion is marked as resolved, I can still see two `outfile.write()` calls in this PR. Can you please check if you have pushed all of the latest commits?

remapping_commands.sh
tskir: Right, but then this would cause problems later in the pipeline because we expect the SNP to always be at 51st position of the read; if the read is shorter the offset will be different. We should probably add a step where all reads shorter than the expected length (101) are filtered out prior to alignment

tskir: (This remains low-priority)

remapping_commands.sh
tskir: Got it. Yes, please add an explanation to the code

reverse_strand.py
mistyskye: Oh I didn't change the else part, this has been fixed.

write_header.py
mistyskye: I've changed how I generate the header, it now just simply adds everything that isn't a contig to my newly generated contigs list to make the new header. If there is still a problem with this, let me know and I will use `bcftools reheader`.

reverse_strand.py
mistyskye: I agree with the biopython addition. I have added it, along with the single `oufile.write`, although I do need an extra check before it so that I don't write the unmapped (flag=4) variants. (see new version when I push it)

reverse_strand.py
mistyskye: I'm not sure how to set the maximum in Sublime Text 3, however I have set `word_wrap true` which adds auto line wrapping to the files.

remapping_commands.sh
mistyskye: This has been discussed and we established that we can't use `bcftools reheader` in its current version as it does not have the functionality we need, so I'll continue using the bash commands to recreate the header.

remapping_commands.sh
mistyskye: I have added this command to the pipeline, although I have left my results calculations as they are a quick way to get the numbers I need to fill in the spreadsheet.

README.md
tskir: @mistyskye Could you please remove the task list from the `README` and move it to the Google doc (if not already there)

reverse_strand.py
tskir: How to set maximum line length in Sublime: https://stackoverflow.com/questions/25900954/80-characters-right-margin-line-in-sublime-text-3 (use 120 instead of 80 as described there)

It looks like the option you currently have enabled wraps the lines to some other limit (or doesn't at all), bercause for example line 157 of the `remapping_commands.sh` is 145 characters long

remapping_commands.sh
mistyskye: I added a filter: all reads not equal to 101 bases are now filtered out.

README.md
```diff
@@ -1,2 +1,90 @@
...
+`chmod a+x remapping_commands.sh reverse_strand.py`
+
+Command usage: (see Input for details)
+`./remapping_commands.sh -g [oldgenome] -n [newgenome] -a [newgenomeaccession] -v [vcffile] -o [outfile name]`
+
+You can use `time` at the beginning of the previous command to get the runtime and CPU time.
+
+**Example command:**
+`time ./remapping_commands.sh -g droso_dm3.fasta -n droso_dm6.fasta -a GCA_000001215.4 -v droso_variants_renamed.vcf \
+	-o test.vcf`
```
tskir: It's better to reformat it to make each flag and its parameter start with the new line, like so:

```bash
time ./remapping_commands.sh \
  -g droso_dm3.fasta \
  -n droso_dm6.fasta \
  -a GCA_000001215.4 \
  -v droso_variants_renamed.vcf \
  -o test.vcf
```

tskir: (It makes the argument list more readable)

remapping_commands.sh
```diff
@@ -0,0 +1,195 @@
+#!/bin/bash
+#requires vcf2bed, bowtie2, samtools, bedtools and reverse_strand.py (uses bamnostic)
+
+#This basic pipeline allows you to input two assemblies and a variant vcf file, 
+#and remap the variants to the new genome
+
+
+#example command:
+#time ./remapping_commands.sh -g droso_dm3.fasta -n GCA_000001215.4_Release_6_plus_ISO1_MT_genomic.fna -a 
+#GCA_000001215.4 -v droso_variants_renamed.vcf -o test.vcf
```
tskir: All comments from the beginning of the file up until this line can be removed because they are already covered in the documentation

remapping_commands.sh
```diff
@@ -0,0 +1,195 @@
...
+		a) newgenomeaccession="${OPTARG}" ;;
+		v) vcffile="${OPTARG}" ;;
+		o) outfile="${OPTARG}" ;;
+	esac
+done
+
+TMPDIR="$(mktemp -d -t --tmpdir=$PWD remapping_tmp_dir_XXX)"
+cd "$TMPDIR"
+echo "-----------------------1) Flanking sequence generation-----------------------"
+#generate chrom sizes:
```
tskir: Minor style comment: it is preferable to leave a space between `#` and the actual comment, and starting the comment with the capital letter, e. g.:

```bash
# Generate chrom sizes
```

Could you please fix this throughout the code?

remapping_commands.sh
```diff
@@ -0,0 +1,195 @@
...
+bowtie2-build "$newgenome" index
+echo "#### Mapping results: ####"
+#bowtie2 arguments:
+#-k 1 means that bowtie2 will only report the best alignment for each variant (default behaviour)
+#-f specifies that the reads are in fasta format
+#-x [index] is the prefix of the index files
+#samtools view arguments:
+#-b: output in BAM format
+#-S: used to specifiy input is SAM format, but only in older versions of samtools
+bowtie2 -k 1 -f -x index "$TMPDIR"/variant_reads.out.fa | samtools view -bS - > "$TMPDIR"/reads_aligned.bam
```
tskir, line 104: It would be helpful for readability to separate different subsections of this section with single empty lines, as you did in the previous section

remapping_commands.sh
```diff
@@ -0,0 +1,195 @@
...
+echo -e "#CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO" >> final_header.txt
+
+#add header to the vcf file:
+cat final_header.txt | cat - var_pre_final.vcf > temp && mv temp vcf_out_with_header.vcf
+#also add it to the ref alleles file so that the line numbers match:
+cat final_header.txt | cat - old_ref_alleles.txt > temp && mv temp old_ref_alleles_with_header.txt
+
+#test each variant to see if REF = ALT, if so, replace the current REF with the corresponding old REF (this is to deal 
+#with variants like G > G)
+.././replace_refs.py -i vcf_out_with_header.vcf -r old_ref_alleles_with_header.txt -o pre_final_vcf.vcf
```
tskir: `.././` is exactly equivalent to `../`, so this can be shortened

remapping_commands.sh
```diff
@@ -0,0 +1,195 @@
...
+total=$(grep "^[^>]" variant_reads.out.fa | wc -l)
+echo $total
+echo "Number of remapped variants:"
+remapped=$(grep "^[^#]" ../"$outfile" | wc -l)
+echo $remapped
+echo "Percentage of remapped variants:"
+percentage=$(bc <<< "scale=3;($remapped/$total)*100")
+echo $percentage"%"
+
+cd ../
```
tskir: In the future we might add the possibility for a user to specify the location of the desired temporary directory. In this case, `cd ../` will not necessarily return us to where we were. It's more reliable to use `cd -` (go back to the previous directory)

replace_refs.py
```diff
@@ -0,0 +1,30 @@
...
+import argparse
+from argparse import RawTextHelpFormatter
+import linecache
+
+description="replaces the REF allele with the old REF allele if REF = ALT in the remapped VCF"
+
+parser = argparse.ArgumentParser(description = description, formatter_class=RawTextHelpFormatter)
+parser.add_argument("-i", "--vcf", help="VCF file containing remapped variants")
+parser.add_argument("-r", "--oldrefalleles", help="file containing the same header as the VCF file + the old REF \
+        alleles in the same order as the VCF")
```
tskir: When splitting string literals, it's not necessary (and actually discouraged) to use the backslash sign. It's better to do it this way:

```python
some_call("some_string_literal_"
          "continuation_of_literal")
```

This will create `some_string_literal_continuation_of_literal`

replace_refs.py
```diff
@@ -0,0 +1,30 @@
...
+                else:
+                        variant=line.split("\t")
+                        if variant[3] == variant[4]:
+                        	#gets corresponding REF allele from the oldrefalleles file
+                                out_vcf.write("%s\t%s\t%s\t%s\t%s\t%s\t%s\t%s" % (variant[0],variant[1],variant[2],
+                                        linecache.getline(args.oldrefalleles, i+1).rstrip(),variant[4],variant[5],
+                                        variant[6],variant[7]))
+                        else:
+                                out_vcf.write(line)
+out_vcf.close()
```
tskir: You can add out_vcf declaration to the context as well:

```python
with open(args.vcf, 'r') as vcf, open(args.outfile, 'w') as out_vcf:
    ...
```

In this case, there'll be no need for closing the file explicitly.

replace_refs.py
```diff
@@ -0,0 +1,30 @@
...
+parser.add_argument("-i", "--vcf", help="VCF file containing remapped variants")
+parser.add_argument("-r", "--oldrefalleles", help="file containing the same header as the VCF file + the old REF \
+        alleles in the same order as the VCF")
+parser.add_argument("-o", "--outfile", help="name of new file")
+args = parser.parse_args()
+
+out_vcf=open(args.outfile, 'w')
+
+with open(args.vcf, 'r') as vcf:
+        for(i, line) in enumerate(vcf):
```
tskir: Our code style (consistent with PEP 8) is to indent using 4 spaces for each level (not 8 spaces as here)

reverse_strand.py
```diff
@@ -0,0 +1,39 @@
+#! /home/juliette/miniconda3/bin/python3
```
tskir: This shebang is not going to be supported on any other machine than your own one; I suggest using the most universal one, `#! /usr/bin/env python3`

reverse_strand.py
```diff
@@ -0,0 +1,39 @@
...
+
+bamfile=pysam.AlignmentFile(args.bam, 'rb')
+outfile=open(args.outfile, 'w')
+old_ref_alleles=open(args.old_ref_alleles, 'w')
+
+#reverses the allele for variants that got mapped onto the reverse strand of the new genome, and prints everything 
+#into correct columns
+
+for read in bamfile:
+		name=read.query_name
```
tskir: Same comment about indentation; needs to be 4 spaces per level

reverse_strand.py
```diff
@@ -0,0 +1,39 @@
...
+		info=name.split("|")
+		nucl=info[3]
+		#mapped onto reverse strand:
+		if read.flag & 16: #decoding bitwise flag with &; may need to be changed back to ==
+			nucl = Seq(nucl, generic_dna).complement()
+		#write it all to the file:
+		if read.flag != 4: #4 means unmapped
```
tskir: This needs to use bitwise operators as well, because the read can be unmapped (4) and also have some other flags set

reverse_strand.py
```diff
@@ -0,0 +1,39 @@
...
+outfile=open(args.outfile, 'w')
+old_ref_alleles=open(args.old_ref_alleles, 'w')
+
+#reverses the allele for variants that got mapped onto the reverse strand of the new genome, and prints everything 
+#into correct columns
+
+for read in bamfile:
+		name=read.query_name
+		info=name.split("|")
+		nucl=info[3]
```
tskir: Code style: space around the assignment operator (`nucl = info[3]`)

reverse_strand.py
```diff
@@ -0,0 +1,39 @@
...
+		#mapped onto reverse strand:
+		if read.flag & 16: #decoding bitwise flag with &; may need to be changed back to ==
+			nucl = Seq(nucl, generic_dna).complement()
+		#write it all to the file:
+		if read.flag != 4: #4 means unmapped
+			outfile.write("%s\t%s\t%s\t%s\t%s\t%s\t%s\n" % (read.reference_name,read.pos+51,info[4],nucl,info[5],
+				info[6],info[7]))
+			#store old reference allele:
```
tskir: Comment style for PEP 8 (here and throughout the code) is to leave space and start with the capital letter, e. g.:
```python
# Store old reference allele
```

tskir (changes requested):
Final comments, mostly on the code style, after this we can merge this PR as the first version of the pipeline is mostly ready

reverse_strand.py
```diff
@@ -0,0 +1,39 @@
...
+
+bamfile=pysam.AlignmentFile(args.bam, 'rb')
+outfile=open(args.outfile, 'w')
+old_ref_alleles=open(args.old_ref_alleles, 'w')
+
+# Reverses the allele for variants that got mapped onto the reverse strand of the new genome, and prints everything 
+# into correct columns
+
+for read in bamfile:
+	name = read.query_name
```
tskir: This file is still using tabs for indentation; it should be 4 spaces instead

reverse_strand.py
```diff
@@ -0,0 +1,39 @@
...
+	info = name.split("|")
+	nucl = info[3]
+	# Mapped onto reverse strand:
+	if read.flag & 16: # Decoding bitwise flag with &; may need to be changed back to ==
+		nucl = Seq(nucl, generic_dna).complement()
+	# Write it all to the file:
+	if not(read.flag & 4): # 4 means unmapped
```
tskir: These can be checked more directly using [`read.is_reverse`](https://pysam.readthedocs.io/en/latest/api.html#pysam.AlignedSegment.is_reverse) and [`read.is_unmapped`](https://pysam.readthedocs.io/en/latest/api.html#pysam.AlignedSegment.is_unmapped) attributes

replace_refs.py
```diff
@@ -0,0 +1,27 @@
+#! /usr/bin/env python3
+import argparse
+from argparse import RawTextHelpFormatter
+import linecache
+
+description="replaces the REF allele with the old REF allele if REF = ALT in the remapped VCF"
+
+parser = argparse.ArgumentParser(description = description, formatter_class=RawTextHelpFormatter)
+parser.add_argument("-i", "--vcf", help="VCF file containing remapped variants")
+parser.add_argument("-r", "--oldrefalleles", help="file containing the same header as the VCF file + the old"
```
tskir:
```suggestion
parser.add_argument("-r", "--oldrefalleles", help="file containing the same header as the VCF file + the old "
```

Space is required, otherwise it will concatenate as "oldREF"

replace_refs.py
```diff
@@ -0,0 +1,27 @@
...
+    for(i, line) in enumerate(vcf):
+        if line.startswith('#'):
+            out_vcf.write(line)
+        else:
+            variant = line.split("\t")
+            if variant[3] == variant[4]:
+            # Gets corresponding REF allele from the oldrefalleles file
+                out_vcf.write("%s\t%s\t%s\t%s\t%s\t%s\t%s\t%s" % (variant[0],variant[1],variant[2],
+                    linecache.getline(args.oldrefalleles, i+1).rstrip(),variant[4],variant[5],
+                    variant[6],variant[7]))
```
tskir: Arguments should be separated with spaces:

```python
variant[0], variant[1], variant[2], ...
```

reverse_strand.py
```diff
@@ -0,0 +1,39 @@
...
+parser = argparse.ArgumentParser(description = description, formatter_class=RawTextHelpFormatter)
+parser.add_argument("-i", "--bam", help="bam file containing remapped variants ")
+parser.add_argument("-o", "--outfile", help="name of new file")
+parser.add_argument("-p", "--old_ref_alleles", help="name of output old ref alleles")
+args = parser.parse_args()
+
+
+bamfile=pysam.AlignmentFile(args.bam, 'rb')
+outfile=open(args.outfile, 'w')
+old_ref_alleles=open(args.old_ref_alleles, 'w')
```
tskir: Space around the assignment operator (here and throughout the code)

README.md
```diff
@@ -1,2 +1,94 @@
...
+
+You can use `time` at the beginning of the previous command to get the runtime and CPU time.
+
+**Example command:**
+`time ./remapping_commands.sh \
+  -g droso_dm3.fasta \
+  -n droso_dm6.fasta \
+  -a GCA_000001215.4 \
+  -v droso_variants_renamed.vcf \
+  -o test.vcf`
```
tskir: I don't think this will properly format because single backticks are only meant to be used for inline code formatting. Here it's better to use block code formatting with triple backticks (as shown here https://help.github.com/en/articles/creating-and-highlighting-code-blocks)

tskir (changes requested):
Some final comments, mostly on code style

tskir (approved):

PR #2, opened 2019-08-09
EVA-1634 — Handle multiple alignments

mistyskye:
- Added flanking sequence length support in reverse_strand.py as a parameter: -f
- Fixed a few errors
- Updated README

### DIFF
--- a/remapping_commands.sh
+++ b/remapping_commands.sh
@@ -2,20 +2,20 @@
 
 oldgenome=''
 newgenome=''
-newgenomeaccession=''
 vcffile=''
 flankingseq=''
 scorecutoff=''
+diffcutoff=''
 outfile=''
 
-while getopts 'g:n:a:v:f:s:o:' flag; do
+while getopts 'g:n:v:f:s:d:o:' flag; do
 	case "${flag}" in
 		g) oldgenome="${OPTARG}";;
 		n) newgenome="${OPTARG}" ;;
-		a) newgenomeaccession="${OPTARG}" ;;
 		v) vcffile="${OPTARG}" ;;
 		f) flankingseq="${OPTARG}" ;;
 		s) scorecutoff="${OPTARG}" ;;
+		d) diffcutoff="${OPTARG}" ;;
 		o) outfile="${OPTARG}" ;;
 	esac
 done
@@ -30,18 +30,19 @@ cut -f1,2 ../"$oldgenome".fai > "$oldgenome".chrom.sizes
 # Filter SNPs only
 bcftools filter -i 'TYPE="snp"' ../"$vcffile" -o snps_only.vcf
 
-# Unnecessary if input file is already SNP-filtered and uniq'ed, but leaving it in as it doesn't change anything:
+# Unnecessary if input file is already SNP-filtered and uniq'ed, uncomment if needed: 
+# (also change the snps_only to snps_only.uniq)
 # Get unique variants based on rsIDs (column 3)
-grep '^#' snps_only.vcf > snps_only.uniq.vcf
-grep -v '^#' snps_only.vcf | sort -u -k3,3  >> snps_only.uniq.vcf
+#grep '^#' snps_only.vcf > snps_only.uniq.vcf
+#grep -v '^#' snps_only.vcf | sort -u -k3,3  >> snps_only.uniq.vcf
 
 # Store header
-bgzip snps_only.uniq.vcf
-bcftools view --header-only snps_only.uniq.vcf.gz -o vcf_header.txt
+bgzip snps_only.vcf
+bcftools view --header-only snps_only.vcf.gz -o vcf_header.txt
 
 # Convert vcf to bed:
-gzip -d snps_only.uniq.vcf.gz
-vcf2bed < snps_only.uniq.vcf > variants.bed
+gzip -d snps_only.vcf.gz
+vcf2bed < snps_only.vcf > variants.bed
 # The actual position of the variant is the second coordinate
 
 # Generate the flanking sequence intervals
@@ -95,7 +96,6 @@ paste <(grep '^>' variants_reads.fa) old_ref_bases.txt variant_bases.txt rsIDs.t
 # >[chr]|[pos interval]|[REF]|[ALT|[rsID]|[QUAL|[FILT]|[INFO]
 # [seq]
 sed 's/\t/|/g; s/ /|/g; s/\(.*\)|/\1 /' temp.txt | tr ' ' '\n' > variant_reads.out.fa
-#total=$(grep "^[^>]" variant_reads.out.fa | wc -l)
 
 echo "---------------------------2) Mapping with bowtie2---------------------------"
 # "index" is the prefix of the output filenames, not a directory
@@ -135,8 +135,8 @@ cd "$TMPDIR"
 samtools sort reads_aligned.bam -o reads_aligned.sorted.bam
 samtools index reads_aligned.sorted.bam
 ../reverse_strand.py -i reads_aligned.sorted.bam -p old_ref_alleles.txt -o variants_remapped.vcf -f "$flankingseq"\
- -s "$scorecutoff"
-shortPerc=$(echo "result = ($shortCount/$total)*100; scale=3; result/1" | bc -l)
+ -s "$scorecutoff" -d "$diffcutoff" 
+shortPerc=$(echo "result = ($shortCount/$total)*100; scale=2; result/1" | bc -l)
 echo "Of the input variants, $shortPerc% were too close to chromosome edges"
 
 # Add interval for variant position in bed format (required by getfasta)
@@ -164,7 +164,7 @@ awk '{print $1}' var_pre_final.vcf | sort | uniq > contig_names.txt
 while read CHR; do echo "##contig=<ID=$CHR>"; done < contig_names.txt > contigs.txt
 
 # Add the reference assembly
-echo "##reference=$newgenomeaccession" >> contigs.txt
+echo "##reference=$newgenome" >> contigs.txt
 
 # Copy everything that isn't #CHROM (the column titles), ##contig or ##reference from the old header to a temp
 awk '($1 !~ /^##contig/ && $1 !~ /^##reference/ && $1 !~ /^#CHROM/) {print $0}' vcf_header.txt > temp_header.txt
--- a/remapping_commands.sh
+++ b/remapping_commands.sh
@@ -4,14 +4,16 @@ oldgenome=''
 newgenome=''
 newgenomeaccession=''
 vcffile=''
+flankingseq=''
 outfile=''
 
-while getopts 'g:n:a:v:o:' flag; do
+while getopts 'g:n:a:v:f:o:' flag; do
 	case "${flag}" in
 		g) oldgenome="${OPTARG}";;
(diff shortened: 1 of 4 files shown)


reverse_strand.py
```diff
@@ -26,14 +28,14 @@
     name = read.query_name
     info = name.split("|")
     nucl = info[3]
+    if read.is_unmapped: # Can be decoded with bitwise flag with & 4 (4 means unmapped)
```
tskir: Minor style comment: per PEP 8 style guide, comment should be separated by two spaces from the code, e. g. :

```python
    if read.is_unmapped:  # Can be decoded with bitwise flag with & 4 (4 means unmapped)
```

(Here and throughout the code)

reverse_strand.py
```diff
@@ -6,18 +6,20 @@
...
 parser.add_argument("-o", "--outfile", help = "name of new file")
 parser.add_argument("-p", "--old_ref_alleles", help = "name of output old ref alleles")
+parser.add_argument("-f", "--flankingseqlength", help="length of each of the flanking sequences")
 args = parser.parse_args()
 
 
 bamfile = pysam.AlignmentFile(args.bam, 'rb')
 outfile = open(args.outfile, 'w')
 old_ref_alleles = open(args.old_ref_alleles, 'w')
+flanklength=args.flankingseqlength
```
tskir: Spaces around the assignment operator

remapping_commands.sh
```diff
@@ -118,7 +121,7 @@ echo '------------------------------3) Data extraction--------------------------
 cd "$TMPDIR"
 samtools sort reads_aligned.bam -o reads_aligned.sorted.bam
 samtools index reads_aligned.sorted.bam
-../reverse_strand.py -i reads_aligned.sorted.bam -p old_ref_alleles.txt -o variants_remapped.vcf
+../reverse_strand.py -i reads_aligned.sorted.bam -p old_ref_alleles.txt -o variants_remapped.vcf -f "$flankingseq"
 
 # Add interval for variant position in bed format (required by getfasta)
 # Reprints all the columns, adding an extra column before the pos column as the pos-1, as bedtools getfasta requires a
```
tskir, line 143: AS/XS filtering is not implemented in the current version of the code in this PR (just adding this discussion in order to not forget it before closing the ticket)

tskir (changes requested):

remapping_commands.sh
```diff
@@ -26,47 +30,57 @@ cut -f1,2 ../"$oldgenome".fai > "$oldgenome".chrom.sizes
 # Filter SNPs only
 bcftools filter -i 'TYPE="snp"' ../"$vcffile" -o snps_only.vcf
 
+# Unnecessary if input file is already SNP-filtered and uniq'ed, but leaving it in as it doesn't change anything:
+# Get unique variants based on rsIDs (column 3)
+grep '^#' snps_only.vcf > snps_only.uniq.vcf
+grep -v '^#' snps_only.vcf | sort -u -k3,3  >> snps_only.uniq.vcf
+
```
tskir, line 38: This step should be removed (or at least commented out, for now), because _in general_ we don't want to change a user's input VCF more than necessary. The variant ID being present multiple lines in a VCF is a problem **for us** because it makes it harder to calculate summary statistics, but theoretically if a user has VCF with repeating identifiers, we should leave them as is. Also consider the possibility that a user's VCF might lack identifiers altogether; this will remove _all but one record_

remapping_commands.sh
```diff
@@ -26,47 +30,57 @@ cut -f1,2 ../"$oldgenome".fai > "$oldgenome".chrom.sizes
...
-bgzip snps_only.vcf
-bcftools view --header-only snps_only.vcf.gz -o vcf_header.txt
+bgzip snps_only.uniq.vcf
+bcftools view --header-only snps_only.uniq.vcf.gz -o vcf_header.txt
 
 # Convert vcf to bed:
-gzip -d snps_only.vcf.gz
-vcf2bed < snps_only.vcf > variants.bed
+gzip -d snps_only.uniq.vcf.gz
+vcf2bed < snps_only.uniq.vcf > variants.bed
```
tskir: After removing/commenting out lines about uniq filtering, these should be changed back as well

remapping_commands.sh
```diff
@@ -81,7 +95,7 @@ paste <(grep '^>' variants_read_filtered.fa) old_ref_bases.txt variant_bases.txt
 # >[chr]|[pos interval]|[REF]|[ALT|[rsID]|[QUAL|[FILT]|[INFO]
 # [seq]
 sed 's/\t/|/g; s/ /|/g; s/\(.*\)|/\1 /' temp.txt | tr ' ' '\n' > variant_reads.out.fa
-
+#total=$(grep "^[^>]" variant_reads.out.fa | wc -l)
```
tskir: Seeing as the line is commented out, is it actually used anywhere?

remapping_commands.sh
```diff
@@ -94,14 +108,16 @@ echo "#### Mapping results: ####"
 
 # Bowtie2 arguments:
 # -k 1 means that bowtie2 will only report the best alignment for each variant (default behaviour)
+# -k 2 means that bowtie2 will report the best alignment and the second best
```
tskir, line 111: Do we actually need the second best alignment itself? I was under the impression that even for `-k 1` mode Bowtie reports an XS score, which is all we need

reverse_strand.py
```diff
@@ -6,18 +6,31 @@
 import pysam
 
 description = "Reverses the allele for variants that got mapped onto the reverse strand of the new genome, and prints"
-            "everything to a new file\n"
+"everything to a new file\n"
```
tskir: This second line would not be included in the description because there are neither backslash not parentheses present; `"everything to a new file\n"` will evaluate as a separate string literal which is not used anywhere. You can use this syntax to make automated literal concatenation work:

```python
description = ("Reverses the allele for variants that got mapped onto the reverse strand of the new genome, and "
               "prints everything to a new file\n")
```

reverse_strand.py
```diff
@@ -26,14 +39,72 @@
...
-    # Write it all to the file:
-    if read.is_unmapped: # Can be decoded with bitwise flag with & 4 (4 means unmapped)
-        outfile.write("%s\t%s\t%s\t%s\t%s\t%s\t%s\n" % (read.reference_name, read.pos+51, info[4], nucl, info[5], 
-            info[6], info[7]))
+    # Getting AS and XS:
+    AS = read.get_tag("AS")
+    try:
+        XS = read.get_tag("XS")  # Some variants don't have secondary alignments, which throws an error
+    except KeyError:
+        XS = -int(flanklength) # Set an arbitrary low value for the "artificial" secondary alignment
```
tskir: PEP8: double space before a `#`

reverse_strand.py
```diff
@@ -26,14 +39,72 @@
...
+    # Getting AS and XS:
+    AS = read.get_tag("AS")
+    try:
+        XS = read.get_tag("XS")  # Some variants don't have secondary alignments, which throws an error
+    except KeyError:
+        XS = -int(flanklength) # Set an arbitrary low value for the "artificial" secondary alignment 
+
+    # Calculating correct position and filtering based on alignement around the variant position
+    start = read.pos
+    readVarPos = start + int(flanklength) + 1
```
tskir: PEP8: camelCase is not used in Python for naming variables; it should be `ready_var_pos` instead (here and throughout the Python modules)

https://www.python.org/dev/peps/pep-0008/#function-and-variable-names

reverse_strand.py
```diff
@@ -26,14 +39,72 @@
...
+    # Create the list:
+    for (cigType, cigLength) in read.cigartuples:
+        for i in range(begin, begin + cigLength):
+            cigList.append(cigType)
+        begin += cigLength
+    for operator in range(0, int(flanklength)+2+localRegionSize):
+        # Stop incrementing varpos once we've reached readVarPos:
+        if start + operator < readVarPos:
+            # Mismatch/matches and deletions increase the position counter:
+            if cigList[operator] == 0 or cigList[operator] == 2:
```
tskir: Instead of 0 and 2, you can use constants `BAM_CMATCH` and `BAM_CDEL` instead

reverse_strand.py
```diff
@@ -26,14 +39,72 @@
...
+    for operator in range(0, int(flanklength)+2+localRegionSize):
+        # Stop incrementing varpos once we've reached readVarPos:
+        if start + operator < readVarPos:
+            # Mismatch/matches and deletions increase the position counter:
+            if cigList[operator] == 0 or cigList[operator] == 2: 
+                varpos += 1
+        # If we are in the local region around the variant, but not at the position of the variant:
+        if (start + operator >= (readVarPos - localRegionSize)) \
+        and (start + operator <= (readVarPos + localRegionSize)) \
+        and (start + operator != (readVarPos)):
```
tskir: As per PEP 8, this should be formatted:

```python
        if (start + operator >= (readVarPos - localRegionSize))
            and (start + operator <= (readVarPos + localRegionSize))
            and (start + operator != (readVarPos)):
```

(Also note that the backslash it not required if you're using parentheses; they create an implied line continuation)

https://www.python.org/dev/peps/pep-0008/#indentation

reverse_strand.py
```diff
@@ -26,14 +39,72 @@
...
+            if cigList[operator] == 0:  # Match or mismatch
+                perfCounter += 1  # Increase the counter for perfect local region
+    if (AS <= int(scorecutoff)):
+        primaryPoorCount += 1
+    elif (AS - XS < 20):
```
tskir: This difference (20) should not be hardcoded and must instead be either a constant or a user-defined parameter. Also, given that AS and XS are both linearly (ish) dependent on flanking sequence length, this should probably be also specified as a fraction of that length

reverse_strand.py
```diff
@@ -26,14 +39,72 @@
...
+old_ref_alleles.close()
+print(totalCount)
+print(unmappedCount)
+print(primaryPoorCount)
+print(gapSmallCount)
+print(contextBadCount)
+print(remappedCount)
+print("% of variants rejected for:")
+print("Unmapped:", round(((unmappedCount/totalCount)*100),3), "%")
+print("Primary alignment too poor:", round(((primaryPoorCount/totalCount)*100), 3), "%")
```
tskir: You don't have to do rounding and formatting manually; there is a very handy string formatting capability in Python (see https://www.learnpython.org/en/String_Formatting for example). In this scenario, you can use something like this:

```python
print("Primary alignment too poor: {:.1%}".format(primaryPoorCount/totalCount))
```

tskir (changes requested):

remapping_commands.sh
mistyskye: [From the docs](http://bowtie-bio.sourceforge.net/bowtie2/manual.shtml#k-mode-search-for-one-or-more-alignments-report-each) it seems that `-k 1` is the default mode whether it's specified or not, [which only reports 1 alignment](http://bowtie-bio.sourceforge.net/manual.shtml#example-4-default--k-1). So I think although it may report an XS, it's always the same as the AS because it doesn't actually look for a 2nd alignment.

tskir: OK, let's just leave it at `-k 2`, it shouldn't do any harm anyway

tskir (approved):

PR #3, opened 2019-08-27
EVA-1588 — Fixed how CIGAR string deletions are handled

mistyskye:
If there are deletions in the read vs the reference, these deletions are now properly taken into account when calculating the variant position

### DIFF
--- a/reverse_strand.py
+++ b/reverse_strand.py
@@ -62,7 +62,6 @@ for read in bamfile:
 
     # Calculating correct position and filtering based on alignement around the variant position
     start = read.pos
-    read_var_pos = start + int(flank_length) + 1
     perf_counter = 0  # This counter will count the number of matches + mismatches around the variant position
     varpos = start  # Store the alignment start position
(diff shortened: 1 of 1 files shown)


reverse_strand.py
```diff
@@ -73,7 +72,9 @@
         for i in range(begin, begin + cig_length):
             cig_list.append(cig_type)
         begin += cig_length
-    for operator in range(0, int(flank_length)+2+local_region_size):
+    # Deletions need to be counted in the new variant position:
+    read_var_pos = start + int(flank_length) + 1 + cig_list.count(pysam.CDEL)
+    for operator in range(0, int(flank_length) + 2 + local_region_size + cig_list.count(pysam.CDEL)):
```
tskir: One minor comment: there is no need to always convert `flank_length` to `int` type, you can specify it in the argument parser:

``` python
parser.add_argument("-f", "--flankingseqlength", type=int, help="length of each of the flanking sequences")
```

Also the same for `score_perc` and `diff_AS_XS_perc` throughout the code

tskir (changes requested):

tskir (approved):

PR #6, opened 2020-04-20
EVA-1597 — Do no skip the indels and process them alongside the SNPs

tcezard:


Evaluating_pipeline.md
```diff
@@ -160,27 +160,29 @@ Hap.py was installed using Singularity and can be accessed by setting `$SINGULAR
...
+    -B ${GENOME_DIR}:/genomes \
     ${SINGULARITY_DIR}/hap.py.simg \
     /opt/hap.py/bin/hap.py  \
         /giab/NA12878/GRCh38/HG001_GRCh38_Fake_Genotypes_shared.vcf.gz \
         /remapping/GRCh37_to_GRCh38/NA12878_f50_s0.6_d0.04_pre_final_vcf_fixed.vcf.gz \
         -f /giab/NA12878/GRCh38/HG001_GRCh38_GIAB_highconf_CG-IllFB-IllGATKHC-Ion-10X-SOLID_CHROM1-X_v.3.3.2_highconf_nosomaticdel_noCENorHET7.bed \
         -r /genomes/GRCh38_no_alt/GCA_000001405.15_GRCh38_no_alt_analysis_set.fna \
-        -o /remapping/GRCh37_to_GRCh38/hap.py_assessement 
+        -o /remapping/GRCh37_to_GRCh38/hap.py_assessement \
+        -V
```
tcezard, line 171: That's to get the assessment VCF output showing detail of each variant and how it was assessed

remapping_commands.sh
```diff
@@ -27,22 +27,21 @@ echo "-----------------------1) Flanking sequence generation--------------------
...
-#grep '^#' snps_only.vcf > snps_only.uniq.vcf
-#grep -v '^#' snps_only.vcf | sort -u -k3,3  >> snps_only.uniq.vcf
+#Decompress if required
+if [ "${vcffile: -7}" == ".vcf.gz" ]
+then
+  cp "$vcffile" input.vcf.gz
+  gzip -d -c "$vcffile" > input.vcf
+else
+  cp "$vcffile" input.vcf
+  bgzip -c "$vcffile" > input.vcf.gz
```
jmmut, line 37: wouldn't it be simpler if handling only the uncompressed (or compressed) copy?

tcezard: The previous version was only handling the uncompressed VCF and recompressing it in the next step. Since I start with compressed VCF I thought I would make both files available

jmmut: that's ok for now but I would like to minimise the redundant data. Keep an eye on which is more useful and maybe in the future we can use only one.

jmmut (approved):

PR #7, opened 2020-05-05
EVA-1958: Prep genome with nextflow

tcezard:
Add genome preparation script that uses nextflow.


genome_preparation.nf
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env nextflow
+
+def helpMessage() {
+    log.info"""
+    Index the provided genome with bowtie and samtools
+    Usage:
+            --fasta                     Fasta file containing the reference genome.
+            --outdir                    Directory where th index will be deployed.
```
jmmut:
```suggestion
            --outdir                    Directory where the index will be deployed.
```

jmmut (approved):

genome_preparation.nf
```diff
@@ -0,0 +1,80 @@
...
+        saveAs: { filename -> file(params.fasta).getName() + filename.replaceFirst(/reference/, "") }
+
+    input:
+        path 'reference' from params.fasta
+
+    output:
+        path "reference.*.bt2"
+
+    """
+    bowtie2-build reference reference
```
jmmut: actually, this command puzzles me. are those "reference reference" files? directories? it looks like they are not variables from the nextflow process...

tcezard: I agree this is a confusing command. See in the [remapping commands](https://github.com/EBIvariation/variant-remapping/blob/master/remapping_commands.sh#L105). 
`reference` here is simply a string for a relative path. Nextflow has copied the input genome  to a temporary directory under the name `reference` following the `input` directive and will move the outputs to their final location following the `publishDir` directive.
I should change the second string to `index` to make it more obvious which is which.

.gitignore
```diff
@@ -0,0 +1 @@
+.idea
```
tskir, line 1: So I guess there are two competing approaches: either ignore `.idea` entirely, or commit it into the repository and share the configuration, run profiles, etc. through it. I guess both ways of doing it are fine. In https://github.com/EBIvariation/vep-mapping-pipeline, I followed the latter approach. Do we want to do this uniformly throughout our repositories, or it doesn't matter too much?

conda.yml
```diff
@@ -0,0 +1,8 @@
+name: variant-remapping
+channels:
+  - bioconda
+  - defaults
+dependencies:
+  - bedtools=2.29.2
+  - bowtie2=2.4.1
+  - samtools=1.4.1-0
```
tskir, line 8: I have slight reservations against using Conda/Bioconda, because I feel it adds yet another layer of abstraction and in combination with a local Python install / PIP / virtualenv can lead to all sorts of weird environment and debugging issues. But admittedly I have very little experience with Conda, so let's hope I just wasn't using it correctly :)

genome_preparation.nf
```diff
@@ -0,0 +1,80 @@
...
+        saveAs: { filename -> file(params.fasta).getName() + filename.replaceFirst(/index/, "") }
+
+    input:
+        path 'reference' from params.fasta
+
+    output:
+        path "index.*.bt2"
+
+    """
+    bowtie2-build reference index
```
tskir: I saw @jmmut's comment about this, but I'm afraid even after the fix the command isn't looking clear. I guess the most confusing part is that `reference` is a variable, but `index` is a literal, serving as a basename for the generated index files. I propose the following:
* Rename `reference` to `genome_fasta`
* Change `index` to `"index"` to indicate this is a literal

genome_preparation.nf
```diff
@@ -0,0 +1,80 @@
...
+        saveAs: { filename -> file(params.fasta).getName() + filename.replaceFirst(/reference/, "") }
+
+    input:
+        path 'reference.fai' from genome_fai
+
+    output:
+        path "reference.chrom.sizes" into chrom_sizes
+
+    """
+    cut -f1,2 reference.fai > reference.chrom.sizes
```
tskir: This is also confusing because both `reference.fai` and `reference.chrom.sizes` look like they may be literals, but in fact the first one is a variable while the second one is a literal. To expand on my previous comment, maybe we could adopt the following convention:
* Variables shouldn't include dots, e. g. `reference_fai`
* Literals should be enclosed by quotes, e. g. `"reference.chrom.sizes"`

genome_preparation.nf
```diff
@@ -0,0 +1,80 @@
...
+
+/*
+ * Index the provided reference genome using bowtie build
+ */
+process bowtieGenomeIndex {
+
+    publishDir params.outdir,
+        overwrite: false,
+        mode: "move",
+        saveAs: { filename -> file(params.fasta).getName() + filename.replaceFirst(/index/, "") }
```
tskir: I wish we didn't have to rename files every time using a closure... Do you think we could accept file basename as a parameter and pass it to all steps explicitly? This would allow us to avoid doing any renames like this, and it will probably mean easier integration with other steps in the future

conda.yml
tskir: By the way, could you also add instructions to install the dependencies and to run this simple Nextflow workflow?

tskir (changes requested):

conda.yml
tcezard: I've not used conda before but have usually heard good things about it. I though that would be a nice way to manage the multiple dependencies that the pipeline has. It will also helps if we want to run the pipeline in travis

genome_preparation.nf
tcezard: Yes those closures are not ideal at all and it took me a while to figure out how this actually works. Your suggestion is sensible and I'll give it a try.

genome_preparation.nf
tcezard: I agree that this. is a bit confusing: 
I'm new to Nextflow but here is my understanding:
The command section is a string (note the three double quotes `"""`) which mean anything in it is just literal unless it is prefixed by a `$` sign (see https://www.nextflow.io/docs/latest/script.html#string-interpolation)
So both `reference` and `index` are literals and this command take a genome file called `reference`  and creates index files called `index.*`
I'm happy to change the name `reference` though.

genome_preparation.nf
tcezard: See my previous comments. I also realised that my use of quotes is inconsistent so I'll fix that.

.gitignore
tcezard: This is because I started working with Pycharm and move to Visual Studio Code because it has a Nextflow plugin. The .idea was getting in my way which is why it made it to the `.gitignore`.

genome_preparation.nf
tcezard: It is much better with your suggestion. thank you.

README_nextflow.md
```diff
@@ -0,0 +1,28 @@
+# Variant-remapping using nextflow 
+
+#### Prerequisites:
+- [Miniconda](https://docs.conda.io/en/latest/miniconda.html)
+- [Nextflow](https://www.nextflow.io/docs/latest/getstarted.html#installation)
+
+
+#### Installation
```
tskir:
```suggestion
# Variant remapping using Nextflow 

## Prerequisites
- [Miniconda](https://docs.conda.io/en/latest/miniconda.html)
- [Nextflow](https://www.nextflow.io/docs/latest/getstarted.html#installation)

## Installation
```

README_nextflow.md
```diff
@@ -0,0 +1,28 @@
...
+```bash
+git clone https://github.com/EBIvariation/variant-remapping.git
+conda env create -f conda.yml
+```
+
+#### Execution 
+First activate the conda environment
```
tskir:
```suggestion
## Execution 
First activate the Conda environment:
```

README_nextflow.md
```diff
@@ -0,0 +1,28 @@
...
+```bash
+conda activate variant-remapping
+```
+
+Test data located in `resources` directory
```
tskir:
```suggestion
Test data is located in the `resources` directory:
```

README_nextflow.md
```diff
@@ -0,0 +1,28 @@
...
+```bash
+nextflow run genome_preparation.nf
+```
+
+Or with real data
```
tskir:
```suggestion
Or with real data:
```

conda.yml
tskir: Sure. Let's try it and see how it works. I'm sure if we manage it properly it should also behave accordingly.

genome_preparation.nf
tskir: I done some testing, and it looks like when you do something like `input: path "genome_fasta" from params.fasta`, it _actually creates the symlink_ "genome_fasta" for the command to use. So this is in a sense "literal", but also in the background a symlink is created.

Similarly, `output: path "${basename}.fai" into genome_fai` makes Nextflow treat `${basename}.fai` as a special value and create output file channel from it.

Sorry for the confusion, I'm also new to Nextflow :) I think we'll need to introduce some minor changes to make everything really consistent. I'll comment on them separately

genome_preparation.nf
```diff
@@ -0,0 +1,80 @@
...
+
+    output:
+        path "$basename.*.bt2"
+
+    """
+    bowtie2-build genome_fasta $basename
+    """
+}
+
+
```
tskir:
```suggestion

```

Examples in the Nextflow docs use one blank line between processes

genome_preparation.nf
```diff
@@ -0,0 +1,80 @@
...
+    """
+}
+
+// Show help message
+if (params.help) exit 0, helpMessage()
+
+params.fasta  = "$baseDir/resources/genome.fa"
+params.outdir = "$baseDir/resources/"
+
+
```
tskir:
```suggestion
// The basename variable contains the name and extension of the input file, e. g. "genome.fa"
```

Standard definition for the basename is "file name without the extension", so I think we should clarify what we mean by it here

genome_preparation.nf
```diff
@@ -0,0 +1,80 @@
...
+    input:
+        path "genome_fasta" from params.fasta
+
+    output:
+        path "$basename.*.bt2"
+
+    """
+    bowtie2-build genome_fasta $basename
+    """
```
tskir, line 38:
```suggestion
    input:
        path genome_fasta from params.fasta

    output:
        path "${basename}.*.bt2"

    """
    bowtie2-build ${genome_fasta} ${basename}
    """
```

I think it's better to define `genome_fasta` as a variable than a literal

genome_preparation.nf
```diff
@@ -0,0 +1,80 @@
...
+
+    input:
+        path "$basename" from params.fasta
+
+    output:
+        path "${basename}.fai" into genome_fai
+
+    """
+    samtools faidx $basename
+    """
```
tskir:
```suggestion
    input:
        path "${basename}" from params.fasta

    output:
        path "${basename}.fai" into genome_fai

    """
    samtools faidx ${basename}
    """
```

Just unifying variable definition style

genome_preparation.nf
tskir: Or, for consistency with other steps, which are all defined relative to "basename", we _could_ change this to

```nextflow
input:
    path "$basename" from params.fasta

[...]

bowtie2-build ${basename} ${basename}
```

But this looks confusing, and we started with something similar.

tskir (changes requested):
Everything's is much more clear and readable now. Just a few style comments. There doesn't appear to be a _completely_ consistent way to define all input/output channels, because for some cases we can specify inputs/outputs directly, and in others we have to rely on the fact that output name depends on the input name (in the case of samtools faidx)

genome_preparation.nf
tskir: The diff makes it hard to see, but this suggested comment is right before the definition of "basename" in the file

genome_preparation.nf
tcezard: Not sure I agree with this. 
In my head at least, literals are less confusing than variables. 
When I search through the work directory, I prefer when the file names are hard coded rather than dependent on the input file. This said I'm fine either way.

tcezard: I would add that the only reason I've used variable as opposed to literal in some output was to avoid the rather complex renaming functions in the `publishDir` directive. 
I guess all this depends on conventions we just need to agree on.

tskir (approved):

PR #8, opened 2020-05-28
EVA-1920 — Rewrite the pipeline using Nextflow

BaronKoy:
- Translation of the `remapping_commands.sh` into nextflow
 - Testing in TravisCI of both pipelines.
 - Add dependency to Conda and separate the tools dependencies from the python libraries dependencies (in `requirements.txt`)
 - Add memory requirement for the memory intensive processes


tcezard (changes requested):
Can you revert the changes to the `README_nextflow.md` and `genome_preparation.nf`. They seem to originate from a error in merging master or something ...

BaronKoy:
I reverted the changes to _genome_preparation.nf_ & _README_nextflow.md_ 
Commit 24bec58

tcezard (approved):

.travis.yml
```diff
@@ -0,0 +1,29 @@
+sudo: required
```
tskir: This is not too much important, but we could make this work without sudo by replacing `sudo ln -s ...` with `export PATH=$PATH:/tmp/nextflow`

.travis.yml
```diff
@@ -0,0 +1,29 @@
...
+
+python:
+  - "3.7"
+install:
+  # Install Nextflow
+  - mkdir /tmp/nextflow && cd /tmp/nextflow
+  - wget -qO- get.nextflow.io | bash
+  - sudo ln -s /tmp/nextflow/nextflow /usr/local/bin/nextflow
+  - cd -
+  # command to install dependencies
```
tskir:
```suggestion
  # Install dependencies
```

requirements.txt
```diff
@@ -0,0 +1,2 @@
+pysam
+biopython
```
tskir, line 2: As far as I understand, Conda can also be used to install both Pysam and Biopython. So if we do want to continue using Conda, maybe we want to do it consistently and install _everything_ using it. Alternatively, we could scrap Conda entirely and install everything using standalone commands/PIP. @tcezard @BaronKoy what are your thoughts?

resources/source_vcf.vcf
```diff
@@ -0,0 +1,5 @@
+##fileformat=VCFv4.2
+#CHROM	POS ID  REF ALT QUAL	FILTER	INFO	FORMAT	HG001
+chr1	48	.	C	A	50	PASS	.	GT	1/1
+chr1	98	.	C	G	50	PASS	.	GT	1/1
+chr1	1078	.	A	G	50	PASS	.	GT	1/1
```
tskir, line 5: Since this file is a _test_ resource, and not something actually required to run the pipeline in production, I'm thinking it might be better to move it to `test/resources/source_vf.vcf`

tests/test_pipeline.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+set -e
```
tskir: Since this is bash, I suggest we go full paranoid and use the [complete](https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/) `set -Eeuxo pipefail` configuration.

tests/test_pipeline.sh
```diff
@@ -0,0 +1,31 @@
...
+set -e 
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+SOURCE_DIR=$(dirname $SCRIPT_DIR)
+
+nextflow run ${SOURCE_DIR}/genome_preparation.nf \
+--fasta ${SOURCE_DIR}/resources/genome.fa \
+--outdir ${SOURCE_DIR}/resources/
+
+
```
tskir:
```suggestion

```

Removing an extra blank line

remapping_pipeline.nf
```diff
@@ -0,0 +1,350 @@
+#!/usr/bin/env nextflow
+
+def helpMessage() {
+    log.info"""
+    Index the provided genome with bowtie and samtools
+    Usage:
+            --oldgenome                     Fasta file containing the reference genome.
+            --outdir                    Directory where the index will be deployed.
+    """
```
tskir, line 18: The help message appears to be the same as for the `genome_preparation.nf`, however, the actual remapping pipeline has different purpose and parameters

remapping_pipeline.nf
```diff
@@ -0,0 +1,350 @@
...
+}
+params.flankingseq = 50
+params.scorecutoff = 0.6
+params.diffcutoff =  0.04
```
tskir, line 25: These parameters should also be added to the help message with their descriptions (they can be found in `reverse_strand.py`)

remapping_pipeline.nf
```diff
@@ -0,0 +1,350 @@
...
+params.outfile = "$baseDir/resources/remap_vcf.vcf"
+
+// Show help message
+if (params.help) exit 0, helpMessage()
```
tskir: This two-line block (Show help message comment + command) should probably be moved to directly after the definition of `helpMessage`

remapping_pipeline.nf
```diff
@@ -0,0 +1,350 @@
...
+params.oldgenome = "$baseDir/resources/genome.fa"
+params.newgenome = "$baseDir/resources/genome.fa"
+params.vcffile = "$baseDir/resources/source_vcf.vcf"
+
+// Index files for both old and new genomes 
+oldgenome_chrom_sizes = file("${params.oldgenome}.chrom.sizes")
+oldgenome_fai = file("${params.oldgenome}.fai")
+newgenome_dir = file(params.newgenome).getParent()
+newgenome_basename = file(params.newgenome).getName()
+newgenome_fai = file("${params.newgenome}.fai")
```
tskir, line 36: Some of the input files for this pipeline are actually the output files of the `genome_preparation.nf` step. As far as I understand, in the Nextflow world the recommended approach is write a single big pipeline inlcuding _all_ of the processing step — in this case, it would be easier to transfer files between steps without the need to define extra input and output files.

@tcezard @BaronKoy What do you think about the idea of joining the two `.nf` files into a single pipeline? Since this is a lot of work, I agree that this should be done as a separate ticket and not as part of this one.

remapping_pipeline.nf
```diff
@@ -0,0 +1,350 @@
...
+
+// basename and basedir of the output file to know how to name the output file
+outfile_basename = file(params.outfile).getName()
+outfile_dir = file(params.outfile).getParent()
+
+
```
tskir, line 42:
```suggestion

```
Nextflow convention is one blank line between each block

remapping_pipeline.nf
```diff
@@ -0,0 +1,350 @@
...
+    tr a-z A-Z < genome_alleles.txt > genome_alleles.fixed.txt
+    
+    # Paste all the information into the correct columns
+    cut -f1-3 variants_remapped.vcf > temp_first_3_columns.txt
+    cut -f4-7 variants_remapped.vcf > temp_last_4_columns.txt
+    paste temp_first_3_columns.txt genome_alleles.fixed.txt temp_last_4_columns.txt > var_pre_final.vcf
+    '''
+}
+
+
```
tskir:
```suggestion

```
Extra blank line

remapping_pipeline.nf
```diff
@@ -0,0 +1,350 @@
...
+process StoreVCFHeader {
+
+    input:
+        path "source_vcf" from params.vcffile
+
+    output:
+        path "vcf_header.txt" into vcf_header
+
+    """
+    bcftools view --header-only source_vcf -o vcf_header.txt
```
tskir: Since `source_vcf` is a literal and not a variable, I think it should be renamed to something like `source.vcf` to properly reflect the file extension.

@tcezard If you remember we had a discussion of literals vs. variables which we started in the previous PR. Looking at this one, I agree that using literals is _much_ more intuitive. So let's establish this as our convention for writing Nextflow.  We just have to remember to preferably use the same literals for the same files, e. g. whenever we use the source VCF, we should always call it `source.vcf` and not using any other modifications.

remapping_pipeline.nf
```diff
@@ -0,0 +1,350 @@
...
+    input:
+        path "source_vcf" from params.vcffile
+
+    output:
+        path "variants.bed" into variants_bed
+
+    """
+    vcf2bed < source_vcf > variants.bed
+    """
+}
```
tskir, line 73: @tcezard @BaronKoy point for discussion — I wonder if we should aim for more compact representation, at least for simple commands? Compare:

```nextflow
process ConvertVCFToBed {

    input:
        path "source_vcf" from params.vcffile

    output:
        path "variants.bed" into variants_bed

    """
    vcf2bed < source_vcf > variants.bed
    """
}
```

And:

```nextflow
process ConvertVCFToBed {
    input: path "source_vcf" from params.vcffile
    output: path "variants.bed" into variants_bed
    "vcf2bed < source_vcf > variants.bed"
}
```

The first one is 12 lines and the second one is 5. Also, the second one feels more easily readable to me. What do you think?

remapping_pipeline.nf
tskir: (If we go with this, `source_vcf` also needs to be renamed to `source.vcf` in all other places throughout the pipeline)

remapping_pipeline.nf
```diff
@@ -0,0 +1,350 @@
...
+
+/*
+ * Convert get the flanking region in bed format.
```
tskir:
```suggestion
 * Based on variants BED, generate the flanking regions BED.
```

remapping_pipeline.nf
```diff
@@ -0,0 +1,350 @@
...
+
+    // Memory required is 5 times the size of the fasta in Bytes or at least 1GB
+    memory  Math.max(file(params.newgenome).size() * 5, 1073741824) + ' B'
+
+    input:  
+        path "variant_reads.fa" from variant_reads_with_info
+        // This will get the directory containing the bowtie index linked in the directory
+        file 'bowtie_index' from newgenome_dir
+
+
```
tskir, line 183:
```suggestion

```

tskir: Extra blank line

remapping_pipeline.nf
```diff
@@ -0,0 +1,350 @@
...
+    '''
+}
+
+/*
+ * Align sequence with bowtie2
+ */
+process alignWithBowtie {
+
+    // Memory required is 5 times the size of the fasta in Bytes or at least 1GB
+    memory  Math.max(file(params.newgenome).size() * 5, 1073741824) + ' B'
```
tskir:
```suggestion
    memory Math.max(file(params.newgenome).size() * 5, 1073741824) + ' B'
```
Extra space

remapping_pipeline.nf
```diff
@@ -0,0 +1,350 @@
...
+    output:
+        path "reads_aligned.bam" into reads_aligned_bam
+
+    """
+    bowtie2 -k 2 --np 0 -f -x bowtie_index/${newgenome_basename} variant_reads.fa | samtools view -bS - > reads_aligned.bam
+    """
+}
+
+/*
+ * Sort the bam file with samtools
```
tskir:
```suggestion
 * Sort the bam file with samtools and index the result.
```

remapping_pipeline.nf
```diff
@@ -0,0 +1,350 @@
...
+    """
+    samtools sort reads_aligned.bam -o reads_aligned.sorted.bam
+    samtools index reads_aligned.sorted.bam
+    """
+}
+
+/*
+ * 
+ */
+process reversStrand {
```
tskir:
```suggestion
process reverseStrand {
```

remapping_pipeline.nf
```diff
@@ -0,0 +1,350 @@
...
+    # Paste all the information into the correct columns
+    cut -f1-3 variants_remapped.vcf > temp_first_3_columns.txt
+    cut -f4-7 variants_remapped.vcf > temp_last_4_columns.txt
+    paste temp_first_3_columns.txt genome_alleles.fixed.txt temp_last_4_columns.txt > var_pre_final.vcf
+    '''
+}
+
+
+/*
+ *
```
tskir:
```suggestion
 * Build the header for the final VCF output file.
```

remapping_pipeline.nf
```diff
@@ -0,0 +1,350 @@
...
+ * 
+ */
+process mergeHeaderAndContentAndFixRefAllele {
+
+    input:
+        path "final_header.txt" from final_header
+        path "var_pre_final.vcf" from var_pre_final
+        path "old_ref_alleles.txt" from old_ref_alleles
+
+
```
tskir:
```suggestion

```
Extra blank line

remapping_pipeline.nf
```diff
@@ -0,0 +1,350 @@
...
+    # Copy everything that isn't #CHROM (the column titles), ##contig or ##reference from the old header to a temp
+    awk '(\$1 !~ /^##contig/ && \$1 !~ /^##reference/ && \$1 !~ /^#CHROM/) {print \$0}' vcf_header.txt > temp_header.txt
+
+    # Add the two headers together and add the column names
+    cat temp_header.txt contigs.txt > final_header.txt
+    echo -e "#CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO" >> final_header.txt
+    """
+}
+
+
```
tskir:
```suggestion

```
Extra blank line

remapping_pipeline.nf
```diff
@@ -0,0 +1,350 @@
...
+/*
+ * 
+ */
+process mergeHeaderAndContentAndFixRefAllele {
```
tskir: I would actually recommend splitting this step into two: (1) merge header and content, (2) fix ref/alt alleles. Even though this second step is small, it is very distinct from the first one and is essentially a postprocessing step which should have a separate process.

remapping_pipeline.nf
```diff
@@ -0,0 +1,350 @@
...
+        mode: "copy"
+
+    input:
+        path "genome.fa" from params.newgenome
+        path "pre_final_vcf.vcf" from pre_final_vcf
+
+    output:
+        path "${outfile_basename}" into output_file        
+    """
+    # Run bcftools norm to swap the REF and ALT alleles if the REF doesn't match the new assembly
```
tskir: I think this comment can be safely removed from here and instead added to the description of this process (in the `/* */` comment above its definition)

remapping_pipeline.nf
```diff
@@ -0,0 +1,350 @@
...
+    publishDir outfile_dir,
+        overwrite: true,
+        mode: "copy"
+
+    input:
+        path "genome.fa" from params.newgenome
+        path "pre_final_vcf.vcf" from pre_final_vcf
+
+    output:
+        path "${outfile_basename}" into output_file
```
tskir: `output_file` is very non-descriptive, it should be renamed to something like `output_vcf` or even `final_output_vcf`

remapping_pipeline.nf
```diff
@@ -0,0 +1,350 @@
...
+        path "${outfile_basename}" into output_file        
+    """
+    # Run bcftools norm to swap the REF and ALT alleles if the REF doesn't match the new assembly
+    bgzip -c pre_final_vcf.vcf > pre_final_vcf.vcf.gz
+    bcftools norm -c ws -f genome.fa -N pre_final_vcf.vcf.gz -o ${outfile_basename} -O v
+    """
+}
+
+/*
+ *
```
tskir:
```suggestion
 * Calculate summary statistics for the final output VCF.
```

tskir (changes requested):
Some general comments:

@BaronKoy I understand this is your first development PR, at least in the last year or so? I must say this is **very** good quality, especially for the first(ish) PR. Good job! Don't mind the large number of comments — this is just because the change itself is big, and most of the comments are trivial fixes of formatting and typos.

Regarding the PR title: we usually use the format "Issue ID — Issue name", e.g. in this case it would be "EVA-1920 — Rewrite the pipeline using Nextflow". Including issue ID in the title will allow JIRA to automatically "notice" this PR and link it directly from the issue.

Regarding the branches: currently this PR is submitted from the `master` branch of your fork into the `master` branch of the origin. This is fine by itself and you don't need to change anything for this PR. However, this may become a problem in the future if you have to work on several code versions simultaneously, since your fork only has one `master` branch. Because of this, when submitting PRs we usually work in topic branches, created specifically for one PR. Those topic branches can be either in your fork, or directly in this repository — I don't think we have a convention established for that yet, so either will be fine.

remapping_pipeline.nf
tcezard: I think you're right. Especially since it is now quite easy to make processes run conditionally. 
I've created a new ticket to merge them back together.

remapping_pipeline.nf
tcezard: Sounds reasonable

remapping_pipeline.nf
tcezard: I actually prefer it with all the extra new lines, but that clearly a matter of opinion.
More importantly I think we should be consistent across processes and the more airy version is more applicable to the general case than the compact one.

remapping_pipeline.nf
tcezard: I think you're right.

requirements.txt
tcezard: I'm actually starting to think that Conda isn't such a good tools after all. We've had a number of issues with conflicting tool dependency. It is forcing us to use specific tool version (such as `samtools 1.9` instead of `samtools 1.10`) which is not good. I still think it has some value for simplifying the setup of the test but I would not advocate we use it on our deployment and potentially ditch it entirely.
I would defer this to a different ticket though.

tskir: In case we're not going to use Conda in production, I think we shouldn't use it for tests as well. But agree this should be done in a separate ticket. Did you already create one? Or I can go ahead and create one as well, just don't want to end up with duplicates.

Since migrating away from Conda is not a huge task, maybe what we could do is have a one (or a small number) of aggregate follow-up ticket(s) with several small changes.

remapping_pipeline.nf
tskir: Linking issue here so it's easier to find — https://www.ebi.ac.uk/panda/jira/browse/EVA-2001

remapping_pipeline.nf
tskir: Actually, on the second look, I now also prefer the first, less condensed option. Don't know what got into me yesterday :) Let's use the first approach consistently for everything.

BaronKoy:
I have moved the test files and made the changes from @tskir review.
After testing with the new changes, the pipeline still runs to completion. Let me know your thoughts (the pipeline.nf help message in particular)

tcezard (approved):
I think it looks good now although there are still a couple of things that could be improved.
I've added them to EVA-2001

requirements.txt
tskir: Will follow up in https://www.ebi.ac.uk/panda/jira/browse/EVA-2001.

resources/source_vcf.vcf
tskir: It looks like this has been fixed partially. I can now see the `test/resources` folder with two files, `genome.fa` and `source_vcf.vcf`. However, in the repository root I can _also_ see the the `resources` folder with the same contents. Looks like it needs to be removed.

remapping_pipeline.nf
tskir: @tcezard This has been resolved but not applied — is this accidental, or you don't agree with the new wording? :)

remapping_pipeline.nf
tskir: @tcezard Same here, resolved but not applied

remapping_pipeline.nf
tskir: @tcezard And also here. Maybe a batch of accepted suggestions didn't get through? GitHub sometimes does those things

remapping_pipeline.nf
```diff
@@ -0,0 +1,372 @@
...
+        overwrite: true,
+        mode: "copy"
+
+    input:
+        path "genome.fa" from params.newgenome
+        path "pre_final_vcf.vcf" from pre_final_vcf
+
+    output:
+        path "${outfile_basename}" into final_output_vcf        
+    """
```
tskir, line 349:
```suggestion

    """
```
Adding a blank line between outputs and the actual step command

tskir (approved):
Very minor comments remaining. Mostly some suggestions which didn't get committed, probably due to GitHub bug. @BaronKoy once you address the remaining comments, feel free to merge this PR without further review from me. And once again: great work!

tskir:
> @BaronKoy once you address the remaining comments, feel free to merge this PR without further review from me.

Just to clarify, it's not only the one new comment which I made just now; there are several existing unresolved conversations above. Prior to merging, make sure to go through them as well, including expanding the hidden conversations.

resources/source_vcf.vcf
tcezard: Yes this will get fix in EVA-2001

PR #9, opened 2020-06-05
EVA-2001 Combine genome prep and execution and refactor

tcezard:
- [x] Combine the genome preparation and the processing steps into one pipeline
 - [x] Fail when required file are not provided
 - [x] More stringent test


BaronKoy (commented):
@tcezard Changes look good.

README_nextflow.md
```diff
@@ -17,12 +17,12 @@ conda activate variant-remapping
 pip install -r requirements.txt
```
tskir, line 17: @tcezard General comment about file organisation: currently, almost everything is located directly in the repository root (except for tests). To make everything more neat, I suggest moving the following files into separate folders:
* *.md → **docs/** (except main README.md)
* Pipeline modules, including *.py, *.sh and *.nf →**src/**, or **pipeline/**, or **remapping-pipeline/**, or other reasonably named folder

This can be done as part of this ticket or in a separate one.

README_nextflow.md
```diff
@@ -17,12 +17,12 @@ conda activate variant-remapping
...
+Run the test script to check that you have all the right dependencies installed properly 
 ```bash
-nextflow run genome_preparation.nf
+tests/test_pipeline.sh
 ```
 
 Or with real data:
 ```bash
-nextflow run genome_preparation.nf --fasta <genome.fa> --outdir <output directory>
+nextflow run remapping_pipeline.nf --oldgenome <genome.fa> --newgenome <new_genome.fa> --vcffile <source.vcf> --outfile  <remap.vcf>
```
tskir:
```suggestion
nextflow run remapping_pipeline.nf \
  --oldgenome <genome.fa> \
  --newgenome <new_genome.fa> \
  --vcffile <source.vcf> \
  --outfile <remap.vcf>
```

remapping_pipeline.nf
```diff
@@ -2,34 +2,45 @@
...
     Parameters:
-            params.vcffile = original vcf file containing variants to be remapped
-            params.oldgenome = source genome file used to discover variants from params.vcffile
-            params.outfile = final vcf containing variants mapped to new genome
-            params.newgenome = new genome which the original vcf file will be mapped against
-            params.flankingseq = The length of the flanking sequences that generates reads
-            params.scorecutoff = Percentage of the flanking sequences that should be used as the alignment Score cut-off threshold
-            params.diffcutoff = Percentage of the flanking sequences that should be used as the AS-XS difference cut-off threshold
+            --flankingseq  The length of the flanking sequences that generates reads (default 50)
+            --scorecutoff  Percentage of the flanking sequences that should be used as the alignment Score cut-off threshold (default 0.6)
```
tskir:
```suggestion
            --scorecutoff  Percentage of the flanking sequences that should be used as the alignment score cut-off threshold (default 0.6)
```

remapping_pipeline.nf
```diff
@@ -39,6 +50,94 @@ newgenome_fai = file("${params.newgenome}.fai")
 outfile_basename = file(params.outfile).getName()
 outfile_dir = file(params.outfile).getParent()
 
+/*
+* Index the new reference genome using bowtie_build
+*/
+if ( !file("${newgenome_dir}/${newgenome_basename}.1.bt2").exists()){
```
tskir:
```suggestion
if (!file("${newgenome_dir}/${newgenome_basename}.1.bt2").exists()){
```

remapping_pipeline.nf
```diff
@@ -79,7 +178,7 @@ process flankingRegionBed {
 
     input:
         path "variants.bed" from variants_bed
-        path "genome.chrom.sizes" from oldgenome_chrom_sizes
+        path "genome.chrom.sizes" from chrom_sizes
```
tskir: Since we are dealing with two genomes here, I would actually suggest leaving the name as "oldgenome_chrom_sizes", even though we only have chrom_sizes for one of them and "oldgenome" is redundant

remapping_pipeline.nf
```diff
@@ -52,7 +151,7 @@ process StoreVCFHeader {
         path "vcf_header.txt" into vcf_header
 
     """
-    bcftools view --header-only source.vcf -o vcf_header.txt
+    bcftools view --header-only source.vcf | grep -v '^##FORMAT' >  vcf_header.txt
```
tskir, line 154: Is there a reason for that, do the `##FORMAT` meta-information lines break something?

remapping_pipeline.nf
```diff
@@ -39,6 +50,94 @@ newgenome_fai = file("${params.newgenome}.fai")
...
+            path "${newgenome_basename}" from params.newgenome
+
+        output:
+            path "${newgenome_basename}.fai" into newgenome_fai
+
+        """
+        samtools faidx ${newgenome_basename}
+        """
+    }
+}
```
tskir, line 122: Currently, samtools indexer processes are duplicated for old and new genome. Looks like this can be solved with [input repeaters](https://www.nextflow.io/docs/latest/process.html#input-repeaters).

If the problem with using input repeaters is conditionals (i.e. we want to check whether the index already exists separately for both genomes), then this conditional can be moved inside the actual command to be executed in this form:

```bash
[ ! -f ${newgenome_basename}.fai && samtools faidx ${newgenome_basename} ]
```

tskir: Additionally, I'd like to point out that relying on preexisting indexes might be dangerous — for example, if someone had their FASTA indexed with samtools version X, and we use samtools version Y in our pipeline, it might cause hard to debug conflicts. So maybe we would be better off always generating index from fresh, as part of the processing pipeline.

tskir (changes requested):
I like this PR a lot, everything is much neater now. Left some comments.

Regarding the last point about Conda replacement: looks like we can mark it done, since we discussed Conda on Friday and decided to give it another chance

remapping_pipeline.nf
tcezard: When the input VCF file has FORMAT/Sample columns they are not carried over to the output (rightly or wrongly, I don't know), but the FORMAT definition in the header was. That made `bcftools norm` seg fault. 
With this change we're at least dealing with this consistently between the header and content.

README_nextflow.md
tcezard: That's a good suggestion, I'll do that.

tcezard: I've reorganised the file in the repository and changed the documentation. 
I've left the nextflow pipeline in the top level directory and renamed it `main.nf` because that's a pattern I've seen use in many nextflow pipelines.

remapping_pipeline.nf
tcezard: Changed

remapping_pipeline.nf
tcezard: Ii've not addressed this particular comment because having bot process in one causes some issue in the output channel. (specifically the order in the output channel needs to be known so that the right `.fai` can be used)
This might be due to my limited understanding of how nextflow works at this time but for now I would rather keep the explicit duplication.
About the second point. There is always a risk about using externally constructed index (with samtools/bowtie or others) but most tools have been pretty good at not changing their indexes between version. So the risk is limited and the saving in time and resources can be big.

README_nextflow.md
tskir: Now it looks great, thanks!

remapping_pipeline.nf
tskir: Ah, I see. Let's leave it as is in the current PR, but I have created a follow-up issue based on your comment: #10

remapping_pipeline.nf
tskir: I agree this is not critical and shouldn't be blocking merging of this PR. However, I still think the refactoring of duplicating code fragments and discussing whether it's always safe to use external indexes still warrant a separate look. I've created a follow-up issue: #11.

README.md
```diff
@@ -4,124 +4,72 @@ Pipeline for remapping VCF variants between two arbitrary assemblies in FASTA fo
 **Method**: creates reads from the flanking sequences of each variant, then maps them to the new assembly using 
 bowtie2.
 
-Currently, it only supports SNPs, and not indels.
+Currently, it only SNPs and short indels but has not been tested with larger or more complex variants
```
tskir:
```suggestion
Currently, it only SNPs and short indels but has not been tested with larger or more complex variants.
```

README.md
```diff
@@ -4,124 +4,72 @@ Pipeline for remapping VCF variants between two arbitrary assemblies in FASTA fo
...
 
-**Prerequisites:**
-- `reverse_strand.py`, included (uses [pysam](https://pysam.readthedocs.io/en/latest/api.html), allows reverse strand 
-allele correction (see Note below for further explanation), and filtering based on bowtie2's alignment score)
-- `replace_refs.py`, included (uses old REF allele in the case of new REF=ALT, these then get swapped (see Note 2))
+## Prerequisites
+To run this pipeline you will need to install and configure [Nextflow](https://www.nextflow.io/docs/latest/getstarted.html#installation). 
+The pipeline uses other software that needs to be downloaded and installed locally. You can obtain them manually or use [Miniconda](https://docs.conda.io/en/latest/miniconda.html)
```
tskir:
```suggestion
The pipeline uses other software that needs to be downloaded and installed locally. You can obtain them manually or use [Miniconda](https://docs.conda.io/en/latest/miniconda.html).
```

tskir (approved):
Everything's been resolved and follow-up issues have been created

PR #12, opened 2021-01-14
force alignment to wait for bt2 index creation if needed

apriltuesday:


main.nf
```diff
@@ -68,11 +68,14 @@ if (!file("${newgenome_dir}/${newgenome_basename}.1.bt2").exists()){
...
         output:
             path "$newgenome_basename.*.bt2" 
+            val "done" into bt2_done
 
         """
         bowtie2-build genome_fasta $newgenome_basename
         """
     }
+} else {
+    bt2_done = Channel.value("done")
```
tskir, line 78: @apriltuesday @tcezard 

I've been thinking about this whole situation for quite some time now, and I'm wondering whether `samtoolsFaidxOld` and `samtoolsFaidxNew` could cause similar race conditions? The problem with both of them, as well as with the `bowtieGenomeIndex`, is that the whole _existence_ of those processes is conditional:

```
if (file doesn't exist) then (define process)
```

So in case the file indeed does not exist, the process is created, but is not properly coupled to the downstream processes through inputs/outputs, causing the race condition. One way to solve it is, indeed, to add state variables like `bt2_done` here. However, it doesn't look like this would the recommended approach, as I couldn't find any examples of conditionally defined _processes_ anywhere in Nextflow documentation.

I was wondering if, instead, we could make the processes permanent and move the if block (checking for file existence) _inside_ them, like shown here: https://www.nextflow.io/docs/latest/process.html#conditional-scripts. Then we could do the normal input/output coupling between processes and avoid having the state variables.

tskir: Then of course, `bowtieGenomeIndex` has an additional problem because it creates not one file but multiple. This could  be solved by removing the `path "$newgenome_basename.*.bt2"` output (since it isn't used anywhere anyway) and replacing it with the output containing the index prefix (`${newgenome_basename}`), which would make more sense for the downstream processes to ingest.

tskir: Oh, actually, thinking about it, `bowtieGenomeIndex` is probably prone to an additional _additional_ problem: if `bowtie2-build` is interrupted halfway through for any reason, it might be possible that the `*.1.bt2` file is created but some of the others are not. So maybe what we want here is to actually use a state **file,** rather than a variable. In this scenario the process would run bowtie2-build, and then _at the end_ create a file (either empty or just with 'done' written in it). Then this file could be used to fix both problems listed above: to make the direct coupling possible, and also to use that file to check that the completed index already exists.

tskir (commented):
I wanted to just take a quick look but ended up thinking way too long about this :) TL;DR—the state variable is fine for now, but I think I found a number of problems with how the processes are organised overall. These should be probably addressed in a separate follow-up issue in the future.

main.nf
apriltuesday: @tskir what you're saying makes sense, I agree that following Nextflow's recommended way of doing conditional scripts will probably save us pain in the long run.  I'm happy to close this PR and make a new ticket for doing this properly for all the processes (rather than merging code only to get rid of it immediately).

tskir: I believe we should still merge this, as it contains an immediate resolution of something which prevents the pipeline from working entirely—and then I agree about creating a new ticket to address this properly. @tcezard What are your thoughts?

tcezard: Couple of comments:
1. First I think we should go ahead and accept the change unless there is a clear reason not to.
2. You're right that this is a race condition being fixed but I disagree that it is due to the conditional process. It is due to the fact that `"$newgenome_basename.*.bt2"` is not stored in a `Channel` used in a subsequent process. The solution proposed actually address that by creating a dummy channel that is being used in `alignWithBowtie` (there are obviously other way of doing the same thing though)
3. The conditional definition of process is not **wrong** for say, although you're right that I have not seen any example of it the documentation. I have seen it used in several [other nextflow workflow](https://github.com/nf-core/smrnaseq/blob/a115116a06ad274b475834316d0193e1b756b249/main.nf#L610) and I think it accurately convey the fact that the index is skipped. the documentation you pointed to shows an example of a process that is run differently based on a parameter.
4. we can't remove `path "$newgenome_basename.*.bt2"` from the output of `bowtieGenomeIndex ` because this is used to move the output index to the `newgenome_dir`. You're right that it is not used by anything else though which is why the race condition occurred. 
5. if `bowtie2-build` build does fail then a non 0 exit status will mark this process as failed and dependency won't be run. If we're worried about the exit status not being reliable enough then we should list all the expected index files so they are all checked. I don't think we need the state file.

tcezard (approved):

main.nf
tskir: Thanks Tim, those are all good comments and things to consider! Some additional thoughts:

1. Agreed. I'll go ahead and merge the PR after posting this comment.
2. I agree that multiple forces are at play here to create race conditions, but a conditional definition still seems to be _one_ of them. For an easier example to demonstrate that:
    - `oldgenome_fai` file is first [defined unconditionally](https://github.com/EBIvariation/variant-remapping/blob/50396845715ed27bcd0705f55a13e1cc04582923/main.nf#L44);
    - next we have a [conditional process definition](https://github.com/EBIvariation/variant-remapping/blob/50396845715ed27bcd0705f55a13e1cc04582923/main.nf#L85) to create this file;
    - and then finally `oldgenome_fai` is used by [subsequent processes](https://github.com/EBIvariation/variant-remapping/blob/50396845715ed27bcd0705f55a13e1cc04582923/main.nf#L133).
    - If the file indeed does not exist, the subsequent processes will not wait for the conditionally defined process because their prerequisite `oldgenome_fai` is _already defined_ there at the top.
3. That's a very good example of conditional definitions, thanks for sharing that! However, the logic there is quite different: the steps are defined conditionally only if they really should not _exist_ given some parameters of the pipeline, and it never depends on the existence of the actual files.

For 4 and 5: you're completely right, I didn't consider those things. So keep the `*.bt2` output and scratch the state file idea.

As a conclusion, I still think it's best if all steps in our pipeline would be defined unconditionally and linked explicitly; and if it's necessary to avoid re-creating indexes, it's easier to check this inside a step (either using the if syntax on the command level, or even something simple like `if [-f index_file] create_index` in the command itself).

tskir (approved):

PR #13, opened 2021-02-02
EVA-2290 - Refactor the reverse_strand.py 

tcezard:
Refactor and add unit tests


sundarvenkata-EBI:
I understand that actions have been set up as part of this PR and will kick into action after merge. But would you be able to run the github action **manually**  just for this PR?

variant_remapping_tools/tests/test_reverse_strand.py
```diff
@@ -0,0 +1,81 @@
+from collections import Counter
+from unittest import TestCase
+from unittest.mock import Mock
+
+import pysam
+from pysam.libcalignedsegment import AlignedSegment
```
sundarvenkata-EBI: This does not look like it is being used...

variant_remapping_tools/reverse_strand.py
```diff
@@ -72,43 +20,116 @@
...
-print("Primary and secondary alignments too close: {:.2%}".format(gap_small_count/total_count))
-print("Local region around variant too poorly aligned: {:.2%}".format(context_bad_count/total_count))
\ No newline at end of file
+    counter['remapped'] += 1
+    return True
+
+
+def process_bam_file(bam_file_path, output_file, old_ref_allele_output_file, flank_length, score_perc, diff_AS_XS_perc):
+    bamfile = pysam.AlignmentFile(bam_file_path, 'rb')
+    outfile = open(output_file, 'w')
```
sundarvenkata-EBI: Suggest using the [with expression](https://docs.python.org/2.5/whatsnew/pep-343.html) to close write file handles in case of errors.

sundarvenkata-EBI (approved):

variant_remapping_tools/reverse_strand.py
tcezard: Didn't want to change anything but I guess that won't change the logic.
Done

variant_remapping_tools/tests/test_reverse_strand.py
tcezard: removed

tcezard:
>I understand that actions have been set up as part of this PR and will kick into action after merge. But would you be able to run the github action manually just for this PR?

Turns out that the github action was being run nut failed because of the yaml syntax. This should now be running properly.

.github/workflows/variant_remapping.yml
```diff
@@ -0,0 +1,40 @@
...
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        # Install Nextflow
+        mkdir /tmp/nextflow && cd /tmp/nextflow
+        wget -qO- get.nextflow.io | bash
+        echo "/tmp/nextflow" >> $GITHUB_PATH
+        cd -
```
tskir: Nextflow is actually [available through Bioconda](https://bioconda.github.io/recipes/nextflow/README.html), so seems like this bit could be replaced by just adding it as a dependency in `conda.yml`. An additional rationale for that is that currently this step is very noisy in the logs and adds hundreds of lines like this:
image

.github/workflows/variant_remapping.yml
```diff
@@ -0,0 +1,40 @@
...
+        # Install dependencies
+        # $CONDA is an environment variable pointing to the root of the miniconda directory
+        $CONDA/bin/conda env update --file conda.yml --name base
+        $CONDA/bin/conda run pip install -r requirements.txt
```
tskir:
```suggestion
        $CONDA/bin/conda run pip -q install -r requirements.txt
```

This will still show errors should they occur, but will reduce the log output overall

requirements.txt
```diff
@@ -1,2 +1,3 @@
 pysam==0.15.4
-biopython==1.77
\ No newline at end of file
+biopython==1.77
+pytest
```
tskir:
```suggestion
pytest==6.2.2
```

Since the other two dependencies are pinned, suggest pinning this one as well

variant_remapping_tools/reverse_strand.py
```diff
@@ -72,43 +20,114 @@
...
+    if read.is_secondary:  # We only want to deal with primary reads
+        return False
+    counter['total'] += 1
+    if read.is_unmapped:  # Can be decoded with bitwise flag with & 4 (4 means unmapped)
+        counter['unmapped'] += 1
+        return False
+    # Getting AS and XS:
+    AS = read.get_tag("AS")
+    try:
+        XS = read.get_tag("XS")  # Some variants don't have secondary alignments, which throws an error
```
tskir: This is a super fine level nitpick, but currently there's a mixture of single and double quotes for string literals—maybe now's the time to fix it, before the module grows and it becomes more cumbersome

tskir: (I think single quotes are better but that's a personal choice. PIP 8 just recommends the choice is consistent throughout the codebase.)

variant_remapping_tools/reverse_strand.py
```diff
@@ -72,43 +20,114 @@
...
-print("Local region around variant too poorly aligned: {:.2%}".format(context_bad_count/total_count))
\ No newline at end of file
+    counter['remapped'] += 1
+    return True
+
+
+def process_bam_file(bam_file_path, output_file, old_ref_allele_output_file, flank_length, score_perc, diff_AS_XS_perc):
+    bamfile = pysam.AlignmentFile(bam_file_path, 'rb')
+
+    # Calculate the score cutoff based on flanking seq length:
```
tskir:
```suggestion
    # Calculate the score cutoff based on flanking seq length
```

variant_remapping_tools/tests/test_reverse_strand.py
```diff
@@ -0,0 +1,80 @@
...
+        read.get_tag = lambda x: tags[x]
+        return read
+
+    def setUp(self) -> None:
+        self.counter = Counter()
+        self.score_cutoff = -20
+        self.diff_cutoff = .06
+
+    def test_calculate_variant_position(self):
+        # 101M: Perfect alignment
```
tskir: I think if you convert comments to one-line docstrings, pytest will be able to automatically use them for test "descriptions" and error messages in case something fails

variant_remapping_tools/tests/test_reverse_strand.py
```diff
@@ -0,0 +1,80 @@
...
+        read = self._mock_read(tags={'AS': -6, 'XS': -6})
+        assert is_read_valid(read, self.counter, 50, self.score_cutoff, self.diff_cutoff) is False
+        assert self.counter == Counter({'total': 1, 'gap_small': 1})
+
+    def test_is_read_invalid_alignment_context(self):
+        read = self._mock_read(cigartuples=((pysam.CMATCH, 51), (pysam.CDEL, 2), (pysam.CMATCH, 50)))
+        assert is_read_valid(read, self.counter, 50, self.score_cutoff, self.diff_cutoff) is False
+        assert self.counter == Counter({'total': 1, 'context_bad': 1})
+
+
```
tskir:
```suggestion

```

tskir (approved):
Just some minor nitpicks. Love the tests BTW, succinct and to the point!

apriltuesday (approved):

.github/workflows/variant_remapping.yml
tcezard: Made the the conda install quiet with `-q` and move the nextflow install to conda as well. 
Good call

.github/workflows/variant_remapping.yml
tcezard: Done

variant_remapping_tools/reverse_strand.py
tcezard: That's true and not difficult to change. This is now done.

variant_remapping_tools/tests/test_reverse_strand.py
tcezard: I did not know that. That a good trick.

PR #14, opened 2021-02-21
EVA-2301 fix insertion of ref alleles for deletions

tcezard:
Introduce a hack to keep original ref allele when it was longer than one base (deletion)
Allow processing of gzip files


main.nf
```diff
@@ -142,13 +142,34 @@ process chromSizes {
...
+        output:
+            path "vcf_file" into vcf_file
+
+        """
+        gunzip -c  source.vcf.gz > vcf_file
+        """
+    }
+}else{
+    vcf_file = params.vcffile
+}
```
tskir: Minor changes:
* Align formatting and process name case with examples in Nextflow documentation;
* Since we're using literals (`path "uncompressed.vcf"`) rather than variables (`file uncompressed_vcf`), I think it makes for better readability when those literals have file extensions.

```suggestion
if( file(params.vcffile).getExtension() == 'gz' ) {
    /*
     * Uncompress VCF file
     */
    process uncompressInputVCF {

        input:
            path "source.vcf.gz" from params.vcffile

        output:
            path "uncompressed.vcf" into vcf_file

        """
        gunzip -c source.vcf.gz > uncompressed.vcf
        """
    }
} else {
    vcf_file = params.vcffile
}
```

variant_remapping_tools/replace_refs.py
```diff
@@ -18,10 +18,12 @@
...
         else:
             variant = line.split("\t")
-            if variant[3] == variant[4]:
+            old_ref = linecache.getline(args.oldrefalleles, i+1).rstrip()
+            # This hack only works for SNP and needs better mechanisms for indels
+            if variant[3] == variant[4] or len(old_ref) != 1:
             # Gets corresponding REF allele from the oldrefalleles file
                 out_vcf.write("%s\t%s\t%s\t%s\t%s\t%s\t%s\t%s" % (variant[0], variant[1], variant[2], 
-                    linecache.getline(args.oldrefalleles, i+1).rstrip(), variant[4], variant[5], 
+                    old_ref, variant[4], variant[5],
```
tskir: I'm not sure I entirely understand how this hack works... Do you suppose it warrants adding a few comments here to make it easier to understand?

tskir: If I remember correctly, the purpose of the original check (`variant[3] == variant[4]`) was to see if the REF and ALT _alleles_ became flipped in the new assembly. E.g. the old variant was **C → A,** but the reference allele in the new assembly is A, so we initially get **A → A** and correct it manually to **C → A.**

However, if we always substitute with the old REF allele in case of deletions and indels, we risk losing the _chain_ flip information. E. g. if the old variant was **AAA → A**, and in the new assembly it became **TTT → T**, here we would incorrectly substitute AAA as the REF.

tskir: Oh, and also, in very rare cases, _both_ the chain and the reference can become flipped, which neither the old nor the new approach would handle.

Which makes me thinking, since getting the allele and chain flips right is a crucial step of remapping, perhaps it's worth revising in depth?

main.nf
tskir: Also, there's the conditional definition thing but we already discussed this

variant_remapping_tools/replace_refs.py
tcezard: That is entirely correct. I've expanded the comment to clarify that this hack cannot stay as is and we need a complete overall of the allele swap mechanism to support indel properly.

main.nf
tcezard: This is now done

variant_remapping_tools/replace_refs.py
```diff
@@ -18,10 +18,16 @@
...
+            old_ref = linecache.getline(args.oldrefalleles, i+1).rstrip()
+            # The reference allele extracted from the new genome in insertReferenceAllele and stored in variant[3]
+            # is only one base long. There test "variant[3] == variant[4]" can only work if the ref and alt have
+            # the same length which won't be the case for indels.
+            # For insertion, we're leaving the single base extracted from the new genome (regardless of the alt)
+            # for deletion, we're using the old reference sequence (regardless of the alt)
+            # TODO: Correct this behaviour to have a consistent treatment of all variants types
+            if variant[3] == variant[4] or len(old_ref) != 1:
+                out_vcf.write("%s\t%s\t%s\t%s\t%s\t%s\t%s\t%s" % (variant[0], variant[1], variant[2], old_ref,
+                                                                  variant[4], variant[5], variant[6], variant[7]))
```
apriltuesday: do you think it's worth writing a test for this, similar to what you did with `reverse_strand.py`, just to document the current behaviour?

main.nf
```diff
@@ -142,13 +142,36 @@ process chromSizes {
...
+process uncompressInputVCF {
+
+    input:
+        path "source.vcf" from params.vcffile
+
+    output:
+        path "uncompressed.vcf" into vcf_file
+
+    script:
+        if ( file(params.vcffile).getExtension() == 'gz' )
```
apriltuesday, line 157: i guess to be consistent with nextflow documentation this would be `if( ... )` but to be honest i would much prefer `if (...)`, to be consistent with java/python/the rest of the world...

variant_remapping_tools/replace_refs.py
tcezard: I have now and it was useful. 
it turns out that the current behaviour for SNP does not make sense either.

main.nf
tcezard: I don't mind either way and I think it is not a bad idea to have formatting differences between languages. This makes it more obvious which language is being used.

variant_remapping_tools/replace_refs.py
```diff
@@ -3,25 +3,35 @@
...
+    args = parser.parse_args()
+
+    with open(args.vcf, 'r') as vcf, open(args.outfile, 'w') as out_vcf:
+        for(i, line) in enumerate(vcf):
+            if line.startswith('#'):
+                out_vcf.write(line)
             else:
-                out_vcf.write(line)
\ No newline at end of file
+                out_vcf.write(process_variant(line, i, args.oldrefalleles))
```
sundarvenkata-EBI: Since the second argument is a line number, I suggest passing i+1 to fit the 1-based definition for line number.

variant_remapping_tools/replace_refs.py
```diff
@@ -3,25 +3,35 @@
...
+    variant = line.split('\t')
+    # Gets corresponding REF allele from the oldrefalleles file.
+    old_ref = linecache.getline(oldrefalleles_file, line_num + 1).rstrip()
+    # The reference allele extracted from the new genome in insertReferenceAllele and stored in variant[3]
+    # is only one base long. There test "variant[3] == variant[4]" can only work if the ref and alt have
+    # the same length which won't be the case for indels.
+    # For insertion, we're leaving the single base extracted from the new genome (regardless of the alt)
+    # for deletion, we're using the old reference sequence (regardless of the alt)
+    # TODO: Correct this behaviour to have a consistent treatment of all variants types
+    if variant[3] == variant[4] or len(old_ref) != 1:
```
sundarvenkata-EBI, line 17: There aren't many of MNVs around and may be we can ignore these during the first pass. But I was wondering what happens in those cases which could satisfy the old ref length criteria of not being equal to 1: 
image.

sundarvenkata-EBI (approved):
Looks good!

variant_remapping_tools/replace_refs.py
tskir: Ooh, I see the point now. That's an excellent explanation in the comments now, thanks!

Do you suppose we should log an issue to address this, either here or in JIRA?

main.nf
tskir, line 157:
```suggestion
        if( file(params.vcffile).getExtension() == 'gz' )
```

While this is indeed not how it's done in Python, and I even agree with @apriltuesday that this is perhaps not the most elegant way to my eyes, I think we should still follow the conventions of the language as they're laid out in the docs/style guide.

main.nf
```diff
@@ -142,13 +142,36 @@ process chromSizes {
...
+            """
+            gunzip -c  source.vcf > uncompressed.vcf
```
```suggestion
            gunzip -c source.vcf > uncompressed.vcf
```

main.nf
```diff
@@ -142,13 +142,36 @@ process chromSizes {
...
+            """
+        else
+            """
+            ln -nfs source.vcf uncompressed.vcf
+            """
+}
+
+
```
```suggestion

```

(I think in all other places we have one blank line between processes.)

variant_remapping_tools/replace_refs.py
tskir: I think MNV is a valid concern, but since, as you mentioned, they're quite rare, I think it's OK to address them in subsequent iterations. If we have/will have a backlog issues for this (I raised this here: https://github.com/EBIvariation/variant-remapping/pull/14#discussion_r581785078), we could document the need for processing MNVs in there as well.

variant_remapping_tools/replace_refs.py
```diff
@@ -3,25 +3,35 @@
...
-            # Gets corresponding REF allele from the oldrefalleles file
-                out_vcf.write("%s\t%s\t%s\t%s\t%s\t%s\t%s\t%s" % (variant[0], variant[1], variant[2], 
-                    linecache.getline(args.oldrefalleles, i+1).rstrip(), variant[4], variant[5], 
-                    variant[6], variant[7]))
+if __name__ == "__main__":
+    description = "replaces the REF allele with the old REF allele if REF = ALT in the remapped VCF"
+
+    parser = argparse.ArgumentParser(description=description, formatter_class=RawTextHelpFormatter)
+    parser.add_argument("-i", "--vcf", help="VCF file containing remapped variants")
+    parser.add_argument("-r", "--oldrefalleles", help="file containing the same header as the VCF file + the old "
```
tskir:
```suggestion
    parser.add_argument("-r", "--oldrefalleles", help="File containing the same header as the VCF file + the old "
```

tskir: Also I notice a mixture of single and double quotes for string literals, but since this script is probably going to be heavily refactored in subsequent iterations, it really doesn't matter for now

variant_remapping_tools/tests/test_replace_refs.py
```diff
@@ -0,0 +1,31 @@
+import os
+from unittest import TestCase
+
+from variant_remapping_tools.replace_refs import process_variant
+
+
+class TestReplaceRef(TestCase):
+
+    def test_process_variant(self):
+        # Reference allele from new genome was inserted in the vcf file and as a result the ref and alt are the same
```
tskir:
```suggestion
        """Reference allele from new genome was inserted in the vcf file and as a result the ref and alt are the same."""
```

variant_remapping_tools/tests/test_replace_refs.py
```diff
@@ -0,0 +1,31 @@
...
+        line = '\t'.join(['chr1', '1', '.', 'G', 'G', '.', '.', '.'])
+        oldrefalleles_file = os.path.join(os.path.dirname(__file__), 'oldrefalleles.txt')
+        line_num = 0
+        # The old ref allele was reinserted
+        assert process_variant(line, line_num, oldrefalleles_file) == 'chr1	1	.	C	G	.	.	.'
+        # Note that I think this is wrong: the old ref allele should have been inserted in the ALT not in REF column
+
+    def test_process_variant_deletion(self):
+        # Reference allele from new genome was inserted in the file which make the deletion disappear
```
tskir:
```suggestion
        """Reference allele from new genome was inserted in the file which make the deletion disappear."""
```

variant_remapping_tools/tests/test_replace_refs.py
```diff
@@ -0,0 +1,31 @@
...
+        line = '\t'.join(['chr2', '1', '.', 'T', 'C', '.', '.', '.'])
+        oldrefalleles_file = os.path.join(os.path.dirname(__file__), 'oldrefalleles.txt')
+        line_num = 1
+        # The old ref allele was reinserted because the second line in oldref file is TAG
+        assert process_variant(line, line_num, oldrefalleles_file) == 'chr2	1	.	TAG	C	.	.	.'
+
+    def test_process_variant_no_change(self):
+        # The bases are different so no changes is applied
```
tskir:
```suggestion
        """The bases are different so no changes is applied."""
```

variant_remapping_tools/tests/test_replace_refs.py
```diff
@@ -0,0 +1,31 @@
...
+class TestReplaceRef(TestCase):
+
+    def test_process_variant(self):
+        # Reference allele from new genome was inserted in the vcf file and as a result the ref and alt are the same
+        line = '\t'.join(['chr1', '1', '.', 'G', 'G', '.', '.', '.'])
+        oldrefalleles_file = os.path.join(os.path.dirname(__file__), 'oldrefalleles.txt')
+        line_num = 0
+        # The old ref allele was reinserted
+        assert process_variant(line, line_num, oldrefalleles_file) == 'chr1	1	.	C	G	.	.	.'
+        # Note that I think this is wrong: the old ref allele should have been inserted in the ALT not in REF column
```
tskir, line 16: We can discuss this in the subsequent issue ;-)

But just to shed some light on this, the rationale was like this. If the original file describes change from A to G, and in the new genome G is the reference allele, we don't want to flip the VCF record to “G to A”, because it's a completely different _change._ Especially if the old VCF had some functional annotations etc. attached to the change _from_ REF _to_ ALT, swapping them could have disastrous results.

tskir (approved):
Left some minor comments, but otherwise looks good! As usual, love the tests, very readable & instructive.

apriltuesday (approved):

variant_remapping_tools/replace_refs.py
tcezard: I created [EVA-2328](https://www.ebi.ac.uk/panda/jira/browse/EVA-2328) to track the correction of the allele swap mechanisms.

variant_remapping_tools/replace_refs.py
tcezard: Done

variant_remapping_tools/replace_refs.py
tcezard: I created EVA-2328 to track the correction of the allele swap mechanisms.

PR #15, opened 2021-03-15
EVA-2328 Process reads to get remapped variants and include tests

andresfsilva:


variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -64,32 +63,55 @@ def is_read_valid(read, counter, flank_length, score_cutoff, diff_cutoff):
...
     counter = Counter()
 
-    # Reverses the allele for variants that got mapped onto the reverse strand of the new genome, and prints everything
-    # into correct columns
-    with open(output_file, 'w') as outfile, open(old_ref_allele_output_file, 'w') as old_ref_alleles:
+    with open(output_file, 'w') as outfile:
         for read in bamfile:
-            if is_read_valid(read, counter, flank_length, score_cutoff, diff_cutoff):
+            varpos, perf_counter, local_region_size = calculate_variant_position(read, flank_length)
+            if is_read_valid(read, counter, flank_length, score_cutoff, diff_cutoff, perf_counter, local_region_size):
```
tcezard: Calculating the variant position before the filtering will fail when encountering unmapped sequence.
I you want to avoid running `calculate_variant_position` twice you'll have to split the filtering function.
Remove the unmapped and secondary first, then calculate the variant position and finally test the alignment around the variant.

main.nf
```diff
@@ -338,70 +338,35 @@ process reverseStrand {
...
     output:
-        path "old_ref_alleles.txt" into old_ref_alleles
         path "variants_remapped.vcf" into variants_remapped
 
     """
     # Ensure that we will use the reverse_strand.py from this repo
-    ${baseDir}/variant_remapping_tools/reverse_strand.py -i reads_aligned.sorted.bam \
-        -p old_ref_alleles.txt -o variants_remapped.vcf -f $params.flankingseq \
-        -s $params.scorecutoff -d $params.diffcutoff
+    ${baseDir}/variant_remapping_tools/reads_to_remapped_variants.py -i reads_aligned.sorted.bam \
```
tskir, line 349: In line with this rename, I suggest also renaming the Nextflow step (`reverseStrand`) and updating its description as well.

tests/test_pipeline.sh
```diff
@@ -5,10 +5,22 @@ set -Eeuo pipefail
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 SOURCE_DIR=$(dirname $SCRIPT_DIR)
 
+# Build the Source VCF
+cat << EOT > "${SCRIPT_DIR}/resources/source.vcf"
+##fileformat=VCFv4.2
```
tskir:
```suggestion
##fileformat=VCFv4.3
```

Since it's the latest version (and it should still be valid against 4.3 as far as I can see)

tests/test_pipeline.sh
```diff
@@ -26,12 +38,14 @@ ls ${SCRIPT_DIR}/resources/remap.vcf \
...
+chr2	1818	.	AAC	T	50	PASS	.
+chr2	2030	.	A	TCC	50	PASS	.
 EOT
 
 # Compare vs the expected VCF
 diff "${SCRIPT_DIR}/resources/expected_remap.vcf" <(grep -v '^##' "${SCRIPT_DIR}/resources/remap.vcf")
 
 # Clean up after the test
-rm -rf work .nextflow* ${SCRIPT_DIR}/resources/*remap.vcf* ${SCRIPT_DIR}/resources/new_genome.fa.* ${SCRIPT_DIR}/resources/genome.fa.fai
+#rm -rf work .nextflow* ${SCRIPT_DIR}/resources/*remap.vcf* ${SCRIPT_DIR}/resources/new_genome.fa.* ${SCRIPT_DIR}/resources/genome.fa.fai
```
tskir: Just to confirm, is it intentional that the clean up is now disabled?

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -64,32 +63,55 @@ def is_read_valid(read, counter, flank_length, score_cutoff, diff_cutoff):
     return True
 
 
-def process_bam_file(bam_file_path, output_file, old_ref_allele_output_file, flank_length, score_perc, diff_AS_XS_perc):
+def fetch_bases(fasta, contig, length, start):
```
tskir, line 78:
```suggestion
def fetch_bases(fasta, contig, length, start):
    """Returns a subsection from a specified FASTA contig. The start coordinate is 1-based."""
```

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -64,32 +63,55 @@ def is_read_valid(read, counter, flank_length, score_cutoff, diff_cutoff):
...
+    zero_base_start = start - 1
+    end = zero_base_start + length
+    new_ref = pysam.FastaFile(fasta).fetch(reference=contig, start=zero_base_start, end=end)
```
tskir: Initialising a `FastaFile` instance every time has got to be very expensive. Could this be rearranged to only initialise once and pass a ready instance into this method?

(In fact, once this is done, this method might not be necessary anymore, as fetching a subsection would just be a one liner.)

variant_remapping_tools/reads_to_remapped_variants.py
tskir: (Might not be necessary anymore due to a subsequent comment)

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -64,32 +63,55 @@ def is_read_valid(read, counter, flank_length, score_cutoff, diff_cutoff):
...
+    return new_ref
+
+
+def calculate_new_alleles(old_ref, new_ref, old_alt, is_reverse_strand):
+    # If reverse strand -> calculate complement
+    old_ref_processed = old_ref
+    old_alt_processed = old_alt
+    if is_reverse_strand:
+        old_ref_processed = Seq(old_ref_processed, generic_dna).complement()
+        old_alt_processed = Seq(old_alt_processed, generic_dna).complement()
```
tskir: Shouldn't this be a **reverse** complement both times? Paging @tcezard and @apriltuesday to double check, maybe I'm missing something.

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -64,32 +63,55 @@ def is_read_valid(read, counter, flank_length, score_cutoff, diff_cutoff):
...
 
-def process_bam_file(bam_file_path, output_file, old_ref_allele_output_file, flank_length, score_perc, diff_AS_XS_perc):
+def fetch_bases(fasta, contig, length, start):
+    zero_base_start = start - 1
+    end = zero_base_start + length
+    new_ref = pysam.FastaFile(fasta).fetch(reference=contig, start=zero_base_start, end=end)
+    return new_ref
+
+
+def calculate_new_alleles(old_ref, new_ref, old_alt, is_reverse_strand):
```
tskir, line 88:
```suggestion
def calculate_new_alleles(old_ref, new_ref, old_alt, is_reverse_strand):
    """Given allele and strand information, calculate the new (ref, alt) pair."""
```

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -64,32 +63,55 @@ def is_read_valid(read, counter, flank_length, score_cutoff, diff_cutoff):
...
-                nucl = info[3]
-                # Mapped onto reverse strand:
-                if read.is_reverse:  # Can be decoded with bitwise flag with & 16
-                    nucl = Seq(nucl, generic_dna).complement()
+                old_ref = info[2]
+                old_alt = info[3]
+                is_reverse_strand = read.is_reverse
+
+                new_ref = fetch_bases(new_genome, read.reference_name, len(old_ref), varpos)
+                new_ref, new_alt = calculate_new_alleles(old_ref, new_ref, old_alt, is_reverse_strand)
```
tskir: This is a matter of personal preference, but since the _object.attribute_ notation is succinct enough in this case, I don't see why it can't just be used as is:

```suggestion

                new_ref = fetch_bases(new_genome, read.reference_name, len(old_ref), varpos)
                new_ref, new_alt = calculate_new_alleles(old_ref, new_ref, old_alt, read.is_reverse)
```

variant_remapping_tools/tests/test_reads_to_remapped_variants.py
```diff
@@ -0,0 +1,70 @@
...
+        fasta_path = '../../tests/resources/new_genome.fa'
+        assert fetch_bases(fasta=fasta_path, contig='chr2', length=2, start=1) == 'CA'
+        assert fetch_bases(fasta=fasta_path, contig='chr2', length=1, start=98) == 'C'
+        assert fetch_bases(fasta=fasta_path, contig='chr2', length=1, start=353) == 'G'
+        assert fetch_bases(fasta=fasta_path, contig='chr2', length=1, start=1078) == 'A'
+        assert fetch_bases(fasta=fasta_path, contig='chr2', length=1, start=1404) == 'G'
+        assert fetch_bases(fasta=fasta_path, contig='chr2', length=3, start=1818) == 'AAC'
+        assert fetch_bases(fasta=fasta_path, contig='chr2', length=1, start=2030) == 'A'
+
+    def test_calculate_new_alleles(self):
```
tskir, line 32: Suggestion to add some cases where ref and/or alt lenghts are greater than 1 to test the behaviour of complement vs. reverse complement as mentioned above.

variant_remapping_tools/tests/test_reads_to_remapped_variants.py
```diff
@@ -0,0 +1,70 @@
...
+        process_bam_file(bamfile, output_file, 50, 0.6, 0.04, fasta_path)
+
+        """
+        Expected:
+        chr2	98	.	C	CG	50	PASS	.
+        chr2	1078	.	A	G	50	PASS	.
+        chr2	1818	.	AAC	T	50	PASS	.
+        chr2	2030	.	A	TCC	50	PASS	.
+        """
+
```
tskir: Again a matter of personal choice, but this seems redundant since the actual expected data is just as readable and located immediately below:

```suggestion
```

variant_remapping_tools/tests/test_reads_to_remapped_variants.py
tskir: Or actually, did we decide that indels with reverse complemented alleles are out of scope for this specific PR? I forget. If this is the case, then please ignore for now :)

tskir (changes requested):
A smashing PR, like this a lot! Cleaning up a lot of old bodged code base and setting up the allele handling in positively the right direction. Left some comments, most are minor.

tskir:
P. S. Also the tests are failing due to the test resource filenames. Perhaps something like this could help: https://github.com/EBIvariation/eva-opentargets/blob/master/vep-mapping-pipeline/repeat_expansion_variants/test/repeat_expansion_pipeline_test.py#L17

tests/test_pipeline.sh
andresfsilva: No, it was not intentional. It was just to retain some files in my machine but forgot to switch it back. Changed, thanks

variant_remapping_tools/reads_to_remapped_variants.py
andresfsilva: I would prefer keeping the `fetch_bases` because of the tests (make sure we are getting the desired bases) and also to handle the 1-base to 0-base there. Let me know what you think.

variant_remapping_tools/reads_to_remapped_variants.py
andresfsilva: I agree, I had made so changes already that did not allow me to commit this as a suggestion but it is changed in the code

variant_remapping_tools/tests/test_reads_to_remapped_variants.py
andresfsilva: I just added some cases for simple insertions and deletions

variant_remapping_tools/reads_to_remapped_variants.py
tskir: Yep, absolutely—now that `FastaFile` is initialised outside, I take absolutely no issue with keeping `fetch_bases` as a separate function.

variant_remapping_tools/reads_to_remapped_variants.py
tskir: Following on today's discussion, this should definitely be a `reverse_complement`. This can be fixed either now or in the future iterations once the general algorithm is implemented; for SNPs it doesn't make a difference anyway.

tskir (approved):

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -64,32 +75,60 @@ def is_read_valid(read, counter, flank_length, score_cutoff, diff_cutoff):
...
+
+
+def calculate_new_alleles(old_ref, new_ref, old_alt, is_reverse_strand):
+    """Given allele and strand information, calculate the new (ref, alt) pair."""
+    # If reverse strand -> calculate complement
+    old_ref_processed = old_ref
+    old_alt_processed = old_alt
+    if is_reverse_strand:
+        old_ref_processed = Seq(old_ref_processed, generic_dna).complement()
+        old_alt_processed = Seq(old_alt_processed, generic_dna).complement()
```
tcezard: As Kirill says it should be reverse_complement()
```suggestion
        old_ref_processed = Seq(old_ref_processed, generic_dna).reverse_complement()
        old_alt_processed = Seq(old_alt_processed, generic_dna).reverse_complement()
```

tcezard (approved):

PR #20, opened 2021-03-18
Filter out alignments with more than 1 hits 

tcezard:
Process the unsorted bam file to group the reads by read name.
Filter out any group with more than 1 alignments


main.nf
```diff
@@ -313,61 +313,60 @@ process alignWithBowtie {
...
     output:
-        path "variants_remapped.vcf" into variants_remapped
+        path "variants_remapped_sorted.vcf" into variants_remapped_sorted
 
     """
-    # Ensure that we will use the reverse_strand.py from this repo
-    ${baseDir}/variant_remapping_tools/reads_to_remapped_variants.py -i reads_aligned.sorted.bam \
-        -o variants_remapped.vcf -f $params.flankingseq \
-        -s $params.scorecutoff -d $params.diffcutoff --newgenome genome.fa
+    cat variants_remapped.vcf | awk '\$1 ~ /^#/ {print \$0;next} {print \$0 | "sort -k1,1 -k2,2n"}' > variants_remapped_sorted.vcf
```
tskir, line 351: Can we use `bcftools sort` here? Unless there is a reason not to
```suggestion
        bgzip variants_remapped.vcf
        tabix variants_remapped.vcf.gz
        bcftools sort -o variants_remapped_sorted.vcf -Ov variants_remapped.vcf.gz
```

main.nf
```diff
@@ -313,61 +313,60 @@ process alignWithBowtie {
...
-        path "reads_aligned.sorted.bam.bai" into bam_index
+        path "variants_remapped.vcf" into variants_remapped
 
     """
-    samtools sort reads_aligned.bam -o reads_aligned.sorted.bam
-    samtools index reads_aligned.sorted.bam
+    # Ensure that we will use the reads_to_remapped_variants.py from this repo
+    ${baseDir}/variant_remapping_tools/reads_to_remapped_variants.py -i reads_aligned.bam \
+        -o variants_remapped.vcf -f $params.flankingseq \
+        -s $params.scorecutoff -d $params.diffcutoff --max_alignment 1 \
```
tskir, line 334: I've read the notes in the JIRA ticket; the precision of 0.999994 is beyond impressive, but the recall slipping to 0.8769 is bothering me a bit.

For the time being, I think it's a very good solution, but I'm sure in the future, if we add more metrics and more sophisticated ways of combining them, we could rescue some of those multiple-alignment variants as well.

tskir: (Increased flanking sequence length could also help, I imagine.)

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -104,33 +103,53 @@ def calculate_new_alleles(old_ref, new_ref, old_alt, is_reverse_strand):
...
+                            old_alt = info[3]
+
+                            new_ref = fetch_bases(fasta, read.reference_name, len(old_ref), varpos)
+                            new_ref, new_alt = calculate_new_alleles(old_ref, new_ref, old_alt, read.is_reverse)
+
+                            outfile.write(
+                                '%s\t%s\t%s\t%s\t%s\t%s\t%s\t%s\n' % (read.reference_name, varpos, info[4], new_ref, new_alt,
+                                                                  info[5], info[6], info[7]))
+            else:
+                counter['Too many alignments'] += 1
```
tskir, line 149: It's probably a matter of personal taste, so you don't need to apply this, but I think it might be better with less indentation:

```suggestion
            if len(read_group) > max_nb_alignment:
                counter['Too many alignments'] += 1
                continue
            counter['total'] += 1
            for read in read_group:
                if not is_read_valid(read, counter, flank_length, score_cutoff, diff_cutoff):
                    continue
                varpos, perf_counter, local_region_size = calculate_variant_position(read, flank_length)
                if not is_local_alignment_perfect(counter, perf_counter, local_region_size):
                    continue

                info = read.query_name.split('|')
                old_ref = info[2]
                old_alt = info[3]
                new_ref = fetch_bases(fasta, read.reference_name, len(old_ref), varpos)
                new_ref, new_alt = calculate_new_alleles(old_ref, new_ref, old_alt, read.is_reverse)

                outfile.write(
                    '%s\t%s\t%s\t%s\t%s\t%s\t%s\t%s\n' % (read.reference_name, varpos, info[4], new_ref, new_alt,
                                                      info[5], info[6], info[7]))
```

tskir (approved):
Near 100% precision is just... wow! Left some minor comments.

main.nf
tcezard: Not in this stage because the header has not been added back yet. but in subsequent stage it would work.

variant_remapping_tools/reads_to_remapped_variants.py
tcezard: This was refactor in subsequent PR so I'll leave it as is but I agree with the suggestion.

main.nf
tcezard: Yes that is completely correct. 
In my test so far with short flanking region it is impossible to know which is the correct alignment when multiple alignment exists. In some cases, the best matching alignment is not the correct one. 
The idea behind this new filter is to ensure that **ALL** variants remapped are correct and try again with longer flanking regions for the ones failing there filter.

Issue #10, opened 2020-06-18
Preserve genotypes and other FORMAT fields


tskir:
When the input VCF file has FORMAT/Sample columns they are not carried over to the output (rightly or wrongly, I don't know), but the FORMAT definition in the header was. That made `bcftools norm` seg fault.

_Originally posted by @tcezard in https://github.com/EBIvariation/variant-remapping/pull/9_

Currently we're removing FORMAT lines, but ideally we should carry everything over.

tskir:
The way of reindexing alleles for the genotypes is described in the [Google Doc](https://docs.google.com/document/d/1ayfEPznvqkS3jKehaYH-jzFta9Rvrhqd_TK_ysT5uS8/) as well.

tcezard:
This was done in #27

PR #21, opened 2021-03-23
EVA-1611 - Minimap2 alignment 

tcezard:
This PR address several issues:
 - Replace bowtie2 with minimap2 for the alignment: minimap2 command is currently optimised for short read alignments
 - Split the left and right flanking regions as two paired reads
 - Implement generic allele swap mechanism
 - Refactor the nextflow pipeline using DSL2 version of Nextflow in preparation for running several alignment rounds.


main.nf
```diff
@@ -1,5 +1,11 @@
 #!/usr/bin/env nextflow
```
tskir, line 1: The tests seem to be still not passing

main.nf
```diff
@@ -50,108 +56,16 @@ newgenome_fai = file("${params.newgenome}.fai")
 outfile_basename = file(params.outfile).getName()
 outfile_dir = file(params.outfile).getParent()
 
-/*
-* Index the new reference genome using bowtie_build
-*/
-if (!file("${newgenome_dir}/${newgenome_basename}.1.bt2").exists()){
-
-    process bowtieGenomeIndex {
```
tskir, line 58: I must say, I'm a bit conflicted about dropping bowtie support entirely... I still think it would have made sense to support a suite of aligners and to be able to switch between them. However, I realise it would have required much more work in terms of postprocessing, and we don't have infinite bandwidth, so I understand why the decision was made

main.nf
```diff
@@ -163,215 +77,63 @@ process uncompressInputVCF {
...
     output:
-        path "variants_remapped.vcf" into variants_remapped
+        path "variants_remapped_sorted.vcf", emit: variants_remapped_sorted
 
     """
-    # Ensure that we will use the reverse_strand.py from this repo
-    ${baseDir}/variant_remapping_tools/reads_to_remapped_variants.py -i reads_aligned.sorted.bam \
-        -o variants_remapped.vcf -f $params.flankingseq \
-        -s $params.scorecutoff -d $params.diffcutoff --newgenome genome.fa
+    cat variants_remapped.vcf | awk '\$1 ~ /^#/ {print \$0;next} {print \$0 | "sort -k1,1 -k2,2n"}' > variants_remapped_sorted.vcf
```
tskir: https://github.com/EBIvariation/variant-remapping/pull/20#discussion_r600113314 applies here as well (I guess it's the same change)

main.nf
```diff
@@ -431,12 +193,36 @@ process calculateStats {
...
+            params.newgenome,
+            prepare_new_genome.out.genome_fai
+        )
+        sortVCF(process_split_reads.out.variants_remapped)
+        buildHeader(sortVCF.out.variants_remapped_sorted, StoreVCFHeader.out.vcf_header)
+        mergeHeaderAndContent(buildHeader.out.final_header, sortVCF.out.variants_remapped_sorted)
+        normalise(mergeHeaderAndContent.out.final_vcf_with_header, params.newgenome)
+        calculateStats(normalise.out.final_output_vcf)
+
+
```
```suggestion
```

main.nf
```diff
@@ -431,12 +193,36 @@ process calculateStats {
...
+            prepare_old_genome.out.genome_fai,
+            prepare_old_genome.out.genome_chrom_sizes,
+            params.newgenome,
+            prepare_new_genome.out.genome_fai
+        )
+        sortVCF(process_split_reads.out.variants_remapped)
+        buildHeader(sortVCF.out.variants_remapped_sorted, StoreVCFHeader.out.vcf_header)
+        mergeHeaderAndContent(buildHeader.out.final_header, sortVCF.out.variants_remapped_sorted)
+        normalise(mergeHeaderAndContent.out.final_vcf_with_header, params.newgenome)
+        calculateStats(normalise.out.final_output_vcf)
```
tskir: Just wanted to clarify, since I didn't work with DSL 2: the steps will still execute in parallel, right? So this is not a _sequence_ of operations, but rather a list of recipes to be executed asynchronously once all the inputs become available?

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -7,75 +7,66 @@
...
+    Resolve the variant definition from the flanking region alignment and old variant definition
+    TODO: Link to algorithm description once public
     """
-    if perf_counter != 2 * local_region_size:
-        counter['context_bad'] += 1
-        return False
-    # Filter out AS's that are too low and XS's that are too close to AS, and those with a non-perfect local alignment:
-    counter['remapped'] += 1
-    return True
+    # grab information from the read name
```
```suggestion
    # Grab information from the read name
```

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -7,75 +7,66 @@
...
+    new_ref = fetch_bases(ref_fasta, left_read.reference_name, left_read.reference_end + 1,
+                          right_read.pos - left_read.reference_end)
+    new_pos = left_read.reference_end + 1
+
+    # 1. Handle reference strand change
+    if not left_read.is_reverse and not right_read.is_reverse:
+        # Forward strand alignment
+        old_ref_conv = old_ref
+        old_alt_conv = old_alts
+
```
```suggestion
```

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -7,75 +7,66 @@
...
+    elif left_read.is_reverse and right_read.is_reverse:
+        # Reverse strand alignment
+        old_ref_conv = reverse_complement(old_ref)
+        old_alt_conv = [reverse_complement(alt) for alt in old_alts]
+    else:
```
tskir, line 40: In addition to strandness checks, I think we also need to ensure the relative orientation is correct. Here are some examples (which would all pass current filters as OK):

Correct
```
...................
  >>>>>>   >>>>>>
  LLLLLL   RRRRRR
```

Incorrect
```
...................
  >>>>>>   >>>>>>
  RRRRRR   LLLLLL
```

```
...................
  >>>>>>
  LLLLLL
    >>>>>>
    RRRRRR

```

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -7,75 +7,66 @@
...
-    counter['remapped'] += 1
-    return True
+    # grab information from the read name
+    name = left_read.query_name
+    info = name.split('|')
+    old_ref = info[2]
+    old_alts = info[3].split(',')
+
+    # Define new ref and new pos
+    new_ref = fetch_bases(ref_fasta, left_read.reference_name, left_read.reference_end + 1,
```
tskir, line 26: I think we also need to assert that `left_read.reference_name` and `right_read.reference_name` are the same. In rare cases, they may be different either due to aligner error, or because of genuine genome reassembly

variant_remapping_tools/reads_to_remapped_variants.py
tskir: Actually, we can get **all** of the three checks (same chromosome, same strand, correct orientation) for free if we perform the mapping in paired end mode. Then we'll be able to just check the [`is_proper_pair`](https://pysam.readthedocs.io/en/latest/api.html#pysam.AlignedSegment.is_proper_pair) flag.

This would also solve the need to pair the reads explicitly, as they will be paired by the aligner and then handled in a single pysam record

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -7,75 +7,66 @@
...
+        old_ref_conv = reverse_complement(old_ref)
+        old_alt_conv = [reverse_complement(alt) for alt in old_alts]
+    else:
+        error_msg = (f'Impossible read configuration: '
+                     f'read1 is_reverse: {left_read.is_reverse}, '
+                     f'read2 is_reverse: {right_read.is_reverse}, '
+                     f'read1 position: {left_read.pos}, '
+                     f'read2 position: {right_read.pos}')
+        print(error_msg)
+        raise ValueError(error_msg)
```
tskir, line 47: I actually imagine that such scenarios (read being not properly paired) will occur from time to time in real life situations, either due to aligner error or genuine genome reassembly, and will not always warrant throwing an exception. Perhaps just counting them as one of the modes of failure to remap would be more appropriate?

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -7,75 +7,66 @@
...
+    # 2. Assign new allele sequences
+    if new_ref == old_ref_conv:
+        new_alts = old_alt_conv
+    elif new_ref in old_alt_conv:
+        old_alt_conv.remove(new_ref)
+        new_alts = old_alt_conv
+        new_alts.append(old_ref_conv)
+    else:
+        new_alts = old_alt_conv
+        new_alts.append(old_ref_conv)
```
tskir, line 58: Just to confirm, we currently strip the genotypes from the output, correct? Because, without reindexing the alleles, this would mess up their order

tskir: I know there's #10 to address this, but perhaps adding a TODO here would also help, just in case

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -85,62 +76,91 @@ def fetch_bases(fasta, contig, length, start):
...
+
+
+def _order_reads(group):
+    """Order read and return the most 5' (smallest coordinates) first."""
+    read1, read2 = group
+    if read1.pos <= read2.pos:
+        return read1, read2
     else:
-        return new_ref, old_alt_processed
+        return read2, read1
```
tskir, line 113: This would not work in a rare situation where the reads align in an incorrect orientation to start with (see comments for `calculate_new_variant_definition`)

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -85,62 +76,91 @@ def fetch_bases(fasta, contig, length, start):
...
+
+
+def pass_basic_filtering(primary_group, secondary_group, counter, filter_align_with_secondary, alignment_score_threshold):
+    """Test if the alignment """
```
tskir: The docstring here is incomplete (and also missing for some of the other newly added functions, e.g. `pass_aligned_filtering`)

variant_remapping_tools/tests/test_reads_to_remapped_variants.py
```diff
@@ -53,22 +42,64 @@ def test_process_bam_file(self):
...
+        left_read = Mock(query_name='chr1|48|CAA|C', reference_name='chr2', pos=1, reference_end=47, is_reverse=False)
+        right_read = Mock(query_name='chr1|48|CAA|C', reference_name='chr2', pos=50, reference_end=110, is_reverse=False)
+        with patch('variant_remapping_tools.reads_to_remapped_variants.fetch_bases', return_value='CAA'):
+            assert calculate_new_variant_definition(left_read, right_read, fasta) == (48, 'CAA', ['C'])
+
+        # Reverse strand alignment for Deletion
+        # REF       AAAAAAAAAAAAAAAAATTGCCCCCCCCCCCCCCCCC
+        #             read2                       read1
+        #    <-----------------------TTG<-----------------------
+        #                            G
```
tskir:
```suggestion
        # Reverse strand alignment for a deletion
        # REF  AAAAAAAAAAAAAAAAATTGCCCCCCCCCCCCCCCCC
        #            read 2             read 1
        #      <----------------TTG<----------------
        #                       G
```

variant_to_realignment.nf
```diff
@@ -0,0 +1,215 @@
+#!/usr/bin/env nextflow
+
+
+// Enable syntax extension
+// See https://www.nextflow.io/docs/latest/dsl2.html
+nextflow.enable.dsl=2
+
+
+/*
+ * Convert VCF file to bed format
```
```suggestion
 * Convert the VCF file to BED format.
```

variant_to_realignment.nf
```diff
@@ -0,0 +1,215 @@
...
+            extractVariantInfoToFastaHeader.out.variant_read1_with_info,
+            extractVariantInfoToFastaHeader.out.variant_read2_with_info,
+            new_genome_fa
+        )
+        readsToRemappedVariants(alignWithMinimap.out.reads_aligned_bam, new_genome_fa)
+
+    emit:
+        variants_remapped = readsToRemappedVariants.out.variants_remapped
+        variants_unmapped = readsToRemappedVariants.out.variants_unmapped
+}
```
tskir, line 239: And a final question, did you run the precision/recall calculation on this pipeline after the changes? I would be interested to see the metrics

tskir (changes requested):
Wow, quite possible the biggest & the most substantial PR in the history of this repo! I left a number of comments, but overall this is a huge step towards a mature, production ready pipeline

main.nf
tcezard: I agree that keeping bowtie2 would be useful at least until we established that minimap2 is performing well (This is not clear yet). 
Eventually I would prefer settling on only one aligner to reduce the maintenance cost though

main.nf
tcezard: I was missing the update of the `conda.yml` but now that I've added them, conda is finding conflict of their dependencies. (Note that is works fine if you install them side by side without conda ...) 
Conda is a mess, [your original comment](https://github.com/EBIvariation/variant-remapping/pull/8#discussion_r433796517) of not using conda  at all sounds good right now  :face_with_head_bandage:

main.nf
tcezard: I think the same rule apply on parallelism as before. All independent tasks are executed in parallel and interdependent tasks are in serial.

variant_to_realignment.nf
tcezard: I did the current state is similar a to what I was getting with bowtie2:

Type | Filter | TRUTH.TOTAL | TRUTH.TP | TRUTH.FN | QUERY.TOTAL | QUERY.FP | QUERY.UNK | FP.gt | METRIC.Recall | METRIC.Precision
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
INDEL | ALL | 383937 | 325483 | 58454 | 325529 | 46 | 0 | 0 | 0.847751 | 0.999859
SNP | ALL | 3254376 | 2685067 | 569309 | 2685075 | 8 | 0 | 0 | 0.825064 | 0.999997

variant_remapping_tools/reads_to_remapped_variants.py
tcezard: This should be covered by the `pass_basic_filtering` function which will filter out any read pair where the reference name are different.

variant_remapping_tools/reads_to_remapped_variants.py
tcezard: These should not occur: 
The `left_read` is defined as having the smallest `pos` of the the two so the only possible occurrences are:
 ```
...................
  >>>>>>   >>>>>>
  LLLLLL   RRRRRR
```
which is correct

```
...................
  >>>>>>
  LLLLLL
    >>>>>>
    RRRRRR

 Or

...................
  >>>>>>
  LLLLLL
    <<<<<<
    RRRRRR
```
Which should trigger the overlapping filter in `pass_aligned_filtering`

 ```
...................
  >>>>>>   <<<<<<
  LLLLLL   RRRRRR

...................
  <<<<<<   >>>>>>
  LLLLLL   RRRRRR
```
Which is what this `else` statement covers.
I guess this latter ones should be covered by the filtering as well instead of breaking the script execution.
So far I have not encountered any.

variant_remapping_tools/reads_to_remapped_variants.py
tcezard: Exactly right.

variant_remapping_tools/reads_to_remapped_variants.py
tcezard: That's a good point, I'll add the TODO

variant_remapping_tools/reads_to_remapped_variants.py
tcezard: This function is simply to initialise the reads' position. It does not make any statement about their orientation.

variant_remapping_tools/reads_to_remapped_variants.py
tcezard: Good point, I'll add them

main.nf
tcezard: Bowtie was reintroduced in a new workflow.

main.nf
tcezard: I moved the sorting of the VCF to a later stage so I could use bcftools.

variant_remapping_tools/reads_to_remapped_variants.py
tcezard: Done

variant_remapping_tools/reads_to_remapped_variants.py
tcezard: Done

main.nf
tcezard: Tests are now passing! 
I had to remove nextflow from the conda install and reinstate it into the github action workflow

variant_remapping_tools/reads_to_remapped_variants.py
tskir: Oh, that's right, it's done _post_ filtering. Sorry, my mistake!

variant_to_realignment.nf
tskir: Thanks! Agreed, seems similar

remapping_commands.sh
```diff
@@ -131,7 +131,7 @@ echo '------------------------------3) Data extraction--------------------------
 # Extract chromosome, rsID, position on new genome, original variant allele, qual, filter and info fields, also 
 # creates old_ref_alleles.txt
 cd "$TMPDIR"
-samtools sort reads_aligned.bam -o reads_aligned.sorted.bam
+samtools sort reads_aligned_test.bam -o reads_aligned.sorted.bam
```
tskir: Unless I misunderstand, isn't this file (`remapping_commands.sh`) the original pipeline which was rewritten into Nextflow and is no longer required? Couldn't we just delete it?

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -85,86 +76,138 @@ def fetch_bases(fasta, contig, length, start):
...
+                        help='bam file containing remapped variants ')
+    parser.add_argument('-o', '--outfile', type=str, required=True,
+                        help='name of new  VCF file')
+    parser.add_argument('--out_failed_file', type=str, required=True,
+                        help='name of the file containing reads that did not align correctly')
+    parser.add_argument('-a', '--alignment_score_threshold', type=int, default=-1,
+                        help='')
+    parser.add_argument('-f', '--filter_align_with_secondary', action='store_true', default=False,
+                        help='Filter out alignment that have one or several secondary alignments.')
+    parser.add_argument('-n', '--newgenome', required=True, help='Fasta file of the target genome')
```
```suggestion
    parser.add_argument('-i', '--bam', type=str, required=True,
                        help='Input BAM file with remapped flanking regions')
    parser.add_argument('-o', '--outfile', type=str, required=True,
                        help='Output VCF file with remapped variants')
    parser.add_argument('--out_failed_file', type=str, required=True,
                        help='Name of the file containing reads that did not align correctly')
    parser.add_argument('-a', '--alignment_score_threshold', type=int, default=-1,
                        help='Minimum alignment score (AS) to consider the read properly mapped')
    parser.add_argument('-f', '--filter_align_with_secondary', action='store_true', default=False,
                        help='Filter out alignments that have one or several secondary alignments.')
    parser.add_argument('-n', '--newgenome', required=True, help='FASTA file of the target genome')
```

variant_to_realignment.nf
```diff
@@ -0,0 +1,274 @@
...
+    cut -f 1,3 flanking_r1.bed > position.txt
+
+    # Store ref bases
+    cut -f 6 flanking_r1.bed > old_ref_bases.txt
+
+    # Store rsIDs
+    cut -f 4 flanking_r1.bed > rsIDs.txt
+
+    # Store variant bases
+    cut -f 7  flanking_r1.bed > variant_bases.txt
```
```suggestion
    cut -f 7 flanking_r1.bed > variant_bases.txt
```

variant_to_realignment.nf
```diff
@@ -0,0 +1,274 @@
...
+    # Replace all spaces and tabs with "|":
+    # >[chr]|[pos interval]|[REF]|[ALT|[rsID]|[QUAL|[FILT]|[INFO]|[seq]
+    # Replace the last "|" with a space (this is between the header and the sequence):
+    # >[chr]|[pos interval]|[REF]|[ALT|[rsID]|[QUAL|[FILT]|[INFO] [seq]
+    # And finally replace the space with a newline:
+    # Output:
+    # >[chr]|[pos interval]|[REF]|[ALT|[rsID]|[QUAL|[FILT]|[INFO]
+    # [seq]
+    sed 's/\\t/|/g; s/ /|/g; s/\\(.*\\)|/\\1 /' temp1.txt | tr ' ' '\\n' > variant_read1.out.fa
+    sed 's/\\t/|/g; s/ /|/g; s/\\(.*\\)|/\\1 /' temp2.txt | tr ' ' '\\n' > variant_read2.out.fa
```
tskir, line 134: I just realised something: the flanking region “reads” that we generate are the same orientation. In other words, we expect them to map in either FF or RR orientation. This kind of goes against the assumption that the paired-end sequencing reads are expected to map FR or RF. It seems to be working fine for a moment, but I wonder if it might trip some aligners under some circumstances.

(I don't think it's worth changing this at this point, just an observation.)

tskir (approved):
Great work Tim. Thank you for preserving Bowtie, really appreciate it!

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -85,86 +76,138 @@ def fetch_bases(fasta, contig, length, start):
...
-    print('Local region around variant too poorly aligned: {:.2%}'.format(counter['context_bad'] / counter['total']))
+                        '%s\t%s\t%s\t%s\t%s\t%s\t%s\t%s\n' % (left_read.reference_name, varpos, info[4], new_ref,
+                                                              ','.join(new_alts), info[5], info[6], info[7]))
+                else:
+                    output_failed_alignment(primary_group, out_failed)
+            else:
+                output_failed_alignment(primary_group, out_failed)
+    for metric in ['Too many alignments', 'Flank unmapped', 'Different chromosomes',
+                   'Poor alignment', 'Overlapping alignment', 'Soft-clipped alignments', 'Remapped']:
+        print(f'{counter.get(metric, 0)} ({counter.get(metric, 0)/counter.get("total"):.2%}) variants rejected for {metric}')
```
apriltuesday: This should be equivalent to the following:
```suggestion
        print(f'{counter[metric]} ({counter[metric]/counter["total"]:.2%}) variants rejected for {metric}')
```

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -85,86 +76,138 @@ def fetch_bases(fasta, contig, length, start):
...
-    print('Primary and secondary alignments too close: {:.2%}'.format(counter['gap_small'] / counter['total']))
-    print('Local region around variant too poorly aligned: {:.2%}'.format(counter['context_bad'] / counter['total']))
+                        '%s\t%s\t%s\t%s\t%s\t%s\t%s\t%s\n' % (left_read.reference_name, varpos, info[4], new_ref,
+                                                              ','.join(new_alts), info[5], info[6], info[7]))
+                else:
+                    output_failed_alignment(primary_group, out_failed)
+            else:
+                output_failed_alignment(primary_group, out_failed)
+    for metric in ['Too many alignments', 'Flank unmapped', 'Different chromosomes',
+                   'Poor alignment', 'Overlapping alignment', 'Soft-clipped alignments', 'Remapped']:
```
apriltuesday: Is this just `counter.keys() - {'total'}`?  This might be slightly easier to interpret...

conda.yml
```diff
@@ -1,12 +1,12 @@
 name: variant-remapping
 channels:
-  - bioconda
   - defaults
+  - conda-forge
+  - bioconda
 dependencies:
   - bedtools=2.29.2
-  - bowtie2=2.4.1
```
tskir, line 7: Oh, one more thing: I suppose for now we shouldn't remove bowtie yet

main.nf
```diff
@@ -431,12 +193,65 @@ process calculateStats {
...
+
+//process_with_minimap
+// Workflow without a name is the default workflow that gets executed when the file is run through nextflow
+workflow {
+    main:
+        prepare_old_genome(params.oldgenome)
+        prepare_new_genome(params.newgenome)
+        uncompressInputVCF(params.vcffile)
+        StoreVCFHeader(uncompressInputVCF.out.vcf_file)
+        process_split_reads(
```
apriltuesday: I know nextflow's "java/groovy with hints of python" thing makes naming conventions impossible, but I would prefer we at least name processes consistently... might be a fool's errand...

remapping_commands.sh
tcezard: I must have changed it inadvertently when refactoring. I guess it can be deleted completely.

prepare_genome.nf
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env nextflow
+
+// Enable syntax extension
+// See https://www.nextflow.io/docs/latest/dsl2.html
+nextflow.enable.dsl=2
+
```
apriltuesday, line 6:
```suggestion
```

variant_to_realignment.nf
```diff
@@ -0,0 +1,274 @@
...
+            new_genome_bowtie_index
+        )
+        readsToRemappedVariants(alignWithBowtie.out.reads_aligned_bam, new_genome_fa)
+
+    emit:
+        variants_remapped = readsToRemappedVariants.out.variants_remapped
+        variants_unmapped = readsToRemappedVariants.out.variants_unmapped
+}
+
+
```
```suggestion
```

apriltuesday (approved):
Left some minor comments but I defer to Kirill and yourself for algorithmic correctness... nice work!

variant_to_realignment.nf
tcezard: Yes that is correct. And I don't think this is a problem at all. Most aligners don't make assumption about orientation anymore only about distances. I know Minimap2 does not seem to care about the orientation

PR #25, opened 2021-03-26
EVA-2354 — Perform multiple alignment rounds

tcezard:
This PR bring in two new round of alignment using Minimap2. The first round uses 50 bases paired end reads, the second 2000 bases and the last uses 50Kb.
This also adds a Yaml file that reports the number of variants that are successfully remapped.

### DIFF
--- a/main.nf
+++ b/main.nf
@@ -95,7 +95,7 @@ process storeVCFHeader {
 }
 
 include { prepare_old_genome; prepare_new_genome; prepare_new_genome_bowtie } from './prepare_genome.nf'
-include { process_split_reads; process_split_reads_with_bowtie } from './variant_to_realignment.nf'
+include { process_split_reads; process_split_reads_mid; process_split_reads_with_bowtie } from './variant_to_realignment.nf'
 
 
 /*
@@ -203,6 +203,19 @@ process calculateStats {
     """
 }
 
+process combineVCF {
+    input:
+        path "variants1.vcf"
+        path "variants2.vcf"
+
+    output:
+        path "merge.vcf", emit: merge_vcf
+
+    """
+    cat variants1.vcf variants2.vcf > merge.vcf
+    """
+}
+
 // Take variants remapped to the new genome and merge them back with the original header and sort the output
 workflow finalise {
     take:
@@ -234,7 +247,16 @@ workflow {
             params.newgenome,
             prepare_new_genome.out.genome_fai
         )
-        finalise(process_split_reads.out.variants_remapped, storeVCFHeader.out.vcf_header)
+        process_split_reads_mid(
+            process_split_reads.out.variants_unmapped,
+            params.oldgenome,
+            prepare_old_genome.out.genome_fai,
+            prepare_old_genome.out.genome_chrom_sizes,
+            params.newgenome,
+            prepare_new_genome.out.genome_fai
+        )
+        combineVCF(process_split_reads.out.variants_remapped, process_split_reads_mid.out.variants_remapped)
+        finalise(combineVCF.out.merge_vcf, storeVCFHeader.out.vcf_header)
 }
 
 //process_with_bowtie
--- a/main.nf
+++ b/main.nf
@@ -157,7 +157,6 @@ process sortVCF {
 
     """
     bgzip variants_remapped.vcf
-    tabix variants_remapped.vcf.gz
     bcftools sort -o variants_remapped_sorted.vcf.gz -Oz variants_remapped.vcf.gz
     """
 }
@@ -221,12 +220,13 @@ workflow finalise {
     take:
         variants_remapped
         vcf_header
+        genome
 
     main:
         buildHeader(variants_remapped, vcf_header)
         mergeHeaderAndContent(buildHeader.out.final_header, variants_remapped)
         sortVCF(mergeHeaderAndContent.out.final_vcf_with_header)
-        normalise(sortVCF.out.variants_remapped_sorted_gz, params.newgenome)
+        normalise(sortVCF.out.variants_remapped_sorted_gz, genome)
         calculateStats(normalise.out.final_output_vcf)
 }
 
@@ -256,7 +256,7 @@ workflow {
             prepare_new_genome.out.genome_fai
         )
         combineVCF(process_split_reads.out.variants_remapped, process_split_reads_mid.out.variants_remapped)
-        finalise(combineVCF.out.merge_vcf, storeVCFHeader.out.vcf_header)
+        finalise(combineVCF.out.merge_vcf, storeVCFHeader.out.vcf_header, params.newgenome)
 }
 
 //process_with_bowtie
@@ -275,5 +275,5 @@ workflow process_with_bowtie {
             prepare_new_genome_bowtie.out.genome_fai,
             prepare_new_genome_bowtie.out.bowtie_indexes
         )
-        finalise(process_split_reads_with_bowtie.out.variants_remapped, storeVCFHeader.out.vcf_header)
+        finalise(process_split_reads_with_bowtie.out.variants_remapped, storeVCFHeader.out.vcf_header, params.newgenome)
 }
--- a/main.nf
+++ b/main.nf
@@ -185,20 +185,20 @@ process normalise {
 /*
  * Create file containing remapping stats
  */
-process calculateStats {
+process outputStats {
 
     publishDir outfile_dir,
         overwrite: true,
         mode: "copy"
 
     input:
-        path outfile_basename
+        path "summary"
 
     output:
-        path "${outfile_basename}.stats"
+        path "${outfile_basename}.yml"
 
     """
-    bcftools stats ${outfile_basename} > ${outfile_basename}.stats
+    ln -s summary "${outfile_basename}.yml"
     """
 }
 
@@ -215,19 +215,33 @@ process combineVCF {
     """
 }
 
+process combineYaml {
+    input:
+        path "round1.yml"
+        path "round2.yml"
+
+    output:
+        path "merge.yml", emit: merge_yml
+
+    """
+    cat round1.yml round2.yml > merge.yml
+    """
+}
+
 // Take variants remapped to the new genome and merge them back with the original header and sort the output
 workflow finalise {
     take:
         variants_remapped
         vcf_header
         genome
+        summary
 
     main:
         buildHeader(variants_remapped, vcf_header)
         mergeHeaderAndContent(buildHeader.out.final_header, variants_remapped)
         sortVCF(mergeHeaderAndContent.out.final_vcf_with_header)
         normalise(sortVCF.out.variants_remapped_sorted_gz, genome)
-        calculateStats(normalise.out.final_output_vcf)
+        outputStats(summary)
 }
 
 
@@ -256,7 +270,8 @@ workflow {
             prepare_new_genome.out.genome_fai
         )
         combineVCF(process_split_reads.out.variants_remapped, process_split_reads_mid.out.variants_remapped)
-        finalise(combineVCF.out.merge_vcf, storeVCFHeader.out.vcf_header, params.newgenome)
+        combineYaml(process_split_reads.out.summary_yml, process_split_reads_mid.out.summary_yml)
+        finalise(combineVCF.out.merge_vcf, storeVCFHeader.out.vcf_header, params.newgenome, combineYaml.out.merge_yml)
 }
 
 //process_with_bowtie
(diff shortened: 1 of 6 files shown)


tcezard:
The current results are quite good for both FN and FP:

Type | Filter | TRUTH.TOTAL | TRUTH.TP | TRUTH.FN | QUERY.TOTAL | QUERY.FP | QUERY.UNK | FP.gt | METRIC.Recall | METRIC.Precision
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
INDEL | ALL | 383937 | 383693 | 244 | 383763 | 70 | 0 | 0 | 0.999364 | 0.999818
SNP | ALL | 3254376 | 3250839 | 3537 | 3250853 | 14 | 0 | 0 | 0.998913 | 0.999996

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -2,9 +2,12 @@
 import argparse
 from argparse import RawTextHelpFormatter
 from collections import Counter
+
+import yaml
 from Bio.Seq import Seq
 from Bio.Alphabet import generic_dna
 import pysam
+from pysam.libcvcf import defaultdict
```
```suggestion
```

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -22,48 +25,48 @@ def calculate_new_variant_definition(left_read, right_read, ref_fasta):
...
 
+    operations = []
     # Define new ref and new pos
     new_ref = fetch_bases(ref_fasta, left_read.reference_name, left_read.reference_end + 1,
-                          right_read.pos - left_read.reference_end)
+                          right_read.reference_start - left_read.reference_end)
     new_pos = left_read.reference_end + 1
 
-    # TODO: All th operation bellow should be recorded so they can be used for subsequent changes of the genotypes.
+    # TODO: All the operation bellow should be recorded so they can be used for subsequent changes of the genotypes.
```
apriltuesday: Is this comment still needed, as we're now including the operations in the `INFO` column?

requirements.txt
```diff
@@ -1,3 +1,4 @@
 pysam==0.15.4
 biopython==1.77
 pytest==6.2.2
+pyyaml==5.4
```
apriltuesday, line 4: ^ can't comment on an empty file but it looks like `remapping_commands.sh` is empty and can be removed.

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -2,9 +2,12 @@
 import argparse
 from argparse import RawTextHelpFormatter
 from collections import Counter
```
```suggestion
from collections import Counter, defaultdict
```

variant_to_realignment.nf
```diff
@@ -229,13 +349,16 @@ workflow process_split_reads {
         alignWithMinimap(
             extractVariantInfoToFastaHeader.out.variant_read1_with_info,
             extractVariantInfoToFastaHeader.out.variant_read2_with_info,
-            new_genome_fa
+            new_genome_fa, flank_length
         )
-        readsToRemappedVariants(alignWithMinimap.out.reads_aligned_bam, new_genome_fa)
+        sortByName(alignWithMinimap.out.reads_aligned_bam)
+        // This last step will use all alignments enven the one that have alternative secondary alignments
```
```suggestion
        // This last step will use all alignments even the ones that have alternative secondary alignments
```

variant_to_realignment.nf
```diff
@@ -229,13 +349,16 @@ workflow process_split_reads {
...
+        readsToRemappedVariants(sortByName.out.reads_aligned_sorted_bam, new_genome_fa, flank_length, false)
```
apriltuesday: These three workflows (`process_split_reads`, `mid` and `long`) are identical except for `flank_length` and this call to `readsToRemappedVariants`, is that right?  I might prefer to extract these as parameters and reuse the code, unless you think they might diverge more substantially in the future.

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -152,33 +178,57 @@ def output_failed_alignment(primary_group, outfile):
...
-            if pass_basic_filtering(primary_group, secondary_group, counter, filter_align_with_secondary,
-                                    alignment_score_threshold):
-                left_read, right_read = _order_reads(primary_group)
+            primary_to_supplementary = link_supplementary(primary_group, supplementary_group)
+            if pass_basic_filtering(primary_group, secondary_group, primary_to_supplementary, counter, filter_align_with_secondary):
+                left_read, right_read = _order_reads(primary_group, primary_to_supplementary)
                 if pass_aligned_filtering(left_read, right_read, counter):
                     counter['Remapped'] += 1
-                    varpos, new_ref, new_alts = calculate_new_variant_definition(left_read, right_read, fasta)
+                    varpos, new_ref, new_alts, ops= calculate_new_variant_definition(left_read, right_read, fasta)
```
```suggestion
                    varpos, new_ref, new_alts, ops = calculate_new_variant_definition(left_read, right_read, fasta)
```

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -22,48 +25,48 @@ def calculate_new_variant_definition(left_read, right_read, ref_fasta):
...
         old_ref_conv = reverse_complement(old_ref)
         old_alt_conv = [reverse_complement(alt) for alt in old_alts]
-    else:
-        error_msg = (f'Impossible read configuration: '
-                     f'read1 is_reverse: {left_read.is_reverse}, '
-                     f'read2 is_reverse: {right_read.is_reverse}, '
-                     f'read1 position: {left_read.pos}, '
-                     f'read2 position: {right_read.pos}')
-        print(error_msg)
-        raise ValueError(error_msg)
```
apriltuesday, line 47: Do we not need to track this case at all anymore?  (I agree with [this comment](https://github.com/EBIvariation/variant-remapping/pull/21#discussion_r600133586) about not raising an exception, maybe I've missed some further discussion on this topic...)

variant_to_realignment.nf
```diff
@@ -193,17 +227,30 @@ process readsToRemappedVariants {
...
+    script:
+    if (filter_align_with_secondary)
     """
     # Ensure that we will use the reads_to_remapped_variants.py from this repo
     ${baseDir}/variant_remapping_tools/reads_to_remapped_variants.py -i reads_aligned.bam \
-        -o variants_remapped.vcf  --filter_align_with_secondary \
-        --newgenome genome.fa --out_failed_file variants_unmapped.vcf
+        -o variants_remapped.vcf  --newgenome genome.fa --out_failed_file variants_unmapped.vcf \
+        --flank_length $flank_length --summary summary.yml --filter_align_with_secondary
+    """
```
apriltuesday: missed this on a previous PR but these should probably be indented

tcezard: Done

requirements.txt
tcezard: It is removed now.

variant_remapping_tools/reads_to_remapped_variants.py
tcezard: This case is now filtered in the `pass_aligned_filtering` function so it does not need to be raised here

variant_to_realignment.nf
tcezard: This is due to a limitation of the Nextflow DSL2.
The signature of a process is defined by the workflow that contains it and the process name. This means that they need to be unique and a process can only be called once per workflow.
The only way to call the same process multiple times is to embed it in separate workflows. Hence the three quasi identical workflow. 
See the warning [here](https://www.nextflow.io/docs/latest/dsl2.html#process-definition) and note [there](https://www.nextflow.io/docs/latest/dsl2.html#workflow-composition)

apriltuesday: Got it, thanks for explaining!  DSL2 is taking some getting used to for me...

apriltuesday (approved):

main.nf
```diff
@@ -186,20 +175,49 @@ process normalise {
...
+    """
+}
+
+process combineVCF {
+    input:
+        path "variants1.vcf"
+        path "variants2.vcf"
+        path "variants3.vcf"
+
+
```
```suggestion

```

variant_remapping_tools/reads_to_remapped_variants.py
tskir: I'd like to file a dissenting opinion: I think it's still worth checking for this condition here, _even if_ it's being already checked somewhere upstream. An additional assert here costs next to nothing, but it provides an additional line of defence in case something breaks upstream or in the tests

variant_to_realignment.nf
```diff
@@ -147,16 +151,46 @@ process alignWithMinimap {
...
         path "reads_aligned.bam", emit: reads_aligned_bam
 
+
+    script:
+    if (flanklength < 500)
+        """
+        # Options used by the 'sr' preset with some modifications:
+        # -O6,16 instead of -O12,32 --> reduce indel cost
+        # -B5 instead of -B10 --> reduce mismatch cost
+        # --end-bonus 20  --> bonus score when the end of the read aligns to mimic global alignment.
```
```suggestion
        # --end-bonus 20 --> bonus score when the end of the read aligns to mimic global alignment.
```

variant_to_realignment.nf
tskir: Would something like this work?
* Create a workflow grouping together all of the repeating steps (`convertVCFToBed`, `flankingRegionBed`, `flankingRegionFasta`, ...);
* In each of the three remaining separate workflows (`process_split_reads`, `process_split_reads_mid`, `process_split_reads_long`), call that single workflow with different parameters instead of calling all of the commands separately.

This would reduce the code duplication significantly, while still adhering to the DSL2 principles of not calling the same workflow twice from within the same workflow.

tskir (changes requested):
Precision & recall metrics look **amazing.** After this PR is merged, I'm tempted to call the pipeline nearly production ready and make a release (either v1.0 or v0.1, depending on the major versioning approach). Yes it still doesn't support genotypes etc., but I think it's mature enough already to start competing with the other existing solutions.

I think code duplication due to DSL2 constraints could still be reduced & left some relevant comments. Otherwise great work and great achiements!

variant_to_realignment.nf
tcezard: This does work indeed. I've refactor following your suggestion

variant_remapping_tools/reads_to_remapped_variants.py
tcezard: Alright, that's fair enough. I've reinserted the `raise`

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -79,41 +89,62 @@ def fetch_bases(fasta, contig, start, length):
...
 
-def _order_reads(group):
-    """Order read and return the most 5' (smallest coordinates) first."""
-    read1, read2 = group
-    if read1.pos <= read2.pos:
+def _order_reads(primary_group, primary_to_supplementary):
+    """
+    Order read and return the most 5' (smallest coordinates) first.
+    if a supplementary read exists and is closer to the other read then it is used in place of the primary
+    """
```
tskir, line 126:
```suggestion
    """Order reads and return the most 5' (smallest coordinates) first.
    If a supplementary read exists and is closer to the other read, then it is used in place of the primary."""
```

```suggestion
    """Order reads and return the most 5' (smallest coordinates) first.
    If a supplementary read exists and is closer to the other read then it is used in place of the primary."""
```

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -193,8 +250,11 @@ def main():
                         help='Output VCF file with remapped variants')
     parser.add_argument('--out_failed_file', type=str, required=True,
                         help='Name of the file containing reads that did not align correctly')
-    parser.add_argument('-a', '--alignment_score_threshold', type=int, default=-1,
-                        help='Minimum alignment score (AS) to consider the read properly mapped')
+    parser.add_argument('--flank_length', type=int, required=True,
+                        help='Length of the flanking region used.')
+    parser.add_argument('--summary', type=str, required=True,
+                        help='Yaml files containing the summary metrics')
```
```suggestion
                        help='YAML files containing the summary metrics')
```

variant_to_realignment.nf
```diff
@@ -96,8 +100,8 @@ process extractVariantInfoToFastaHeader {
     // Disable the string interpolation using single quotes
     // https://www.nextflow.io/docs/latest/script.html#string-interpolation
     '''
-    # Store variant positions
-    cut -f 1,3 flanking_r1.bed > position.txt
+    # Store variant positions (Add + 1 to revert to one based position)
```
```suggestion
    # Store variant positions (add + 1 to revert to one based position)
```

tskir (approved):
Looks great!

Issue #24, opened 2021-03-26
Implement local realignment of old REF / new REF / ALT sequences


tskir:
As described at the end of the [Google document](https://docs.google.com/document/d/1ayfEPznvqkS3jKehaYH-jzFta9Rvrhqd_TK_ysT5uS8/edit#heading=h.86r1uc7xuzuf)

tskir:
See also #18

PR #26, opened 2021-04-26
EVA-2380 - Reduce FP

tcezard:
This PR reduces the number of False positive by removing the variants that are remapped with significant changes. 
In testing, we saw that these groups are enriched in False positives and thus are more likely to be errors.

It also fixes a bug in the way the Soft clipping was detected

### DIFF
--- a/variant_remapping_tools/reads_to_remapped_variants.py
+++ b/variant_remapping_tools/reads_to_remapped_variants.py
@@ -19,7 +19,7 @@ def calculate_new_variant_definition(left_read, right_read, ref_fasta):
     TODO: Link to algorithm description once public
     """
     # Flag to highlight low confidence in an event detected
-    low_confidence = False
+    failure_reason = None
 
     # Grab information from the read name
     name = left_read.query_name
@@ -63,13 +63,13 @@ def calculate_new_variant_definition(left_read, right_read, ref_fasta):
         new_alts.append(old_ref_conv)
         operations.append('rac=' + old_ref_conv + '-' + new_ref)
         if len(old_ref_conv) != len(new_ref):
-            low_confidence = True
+            failure_reason = 'Reference Allele length change'
     else:
         new_alts = old_alt_conv
         new_alts.append(old_ref_conv)
         operations.append('rac=' + old_ref_conv + '-' + new_ref)
         operations.append('nra')
-        low_confidence = True
+        failure_reason = 'Novel Reference Allele'
 
     # 3. Correct zero-length reference sequence
     if len(new_ref) == 0:
@@ -78,7 +78,7 @@ def calculate_new_variant_definition(left_read, right_read, ref_fasta):
         new_alts = [new_ref + alt for alt in new_alts]
         operations.append('zlr')
 
-    return new_pos, new_ref, new_alts, operations, low_confidence
+    return new_pos, new_ref, new_alts, operations, failure_reason
 
 
 def fetch_bases(fasta, contig, start, length):
@@ -124,7 +124,7 @@ def group_reads(bam_file_path):
             yield primary_group, supplementary_group, secondary_group
 
 
-def _order_reads(primary_group, primary_to_supplementary):
+def order_reads(primary_group, primary_to_supplementary):
     """
     Order read and return the most 5' (smallest coordinates) first.
     if a supplementary read exists and is closer to the other read then it is used in place of the primary
@@ -225,11 +225,11 @@ def process_bam_file(bam_file_path, output_file, out_failed_file, new_genome, fi
             counter['total'] += 1
             primary_to_supplementary = link_supplementary(primary_group, supplementary_group)
             if pass_basic_filtering(primary_group, secondary_group, primary_to_supplementary, counter, filter_align_with_secondary):
-                left_read, right_read = _order_reads(primary_group, primary_to_supplementary)
+                left_read, right_read = order_reads(primary_group, primary_to_supplementary)
                 if pass_aligned_filtering(left_read, right_read, counter):
-                    varpos, new_ref, new_alts, ops, low_confidence = \
+                    varpos, new_ref, new_alts, ops, failure_reason = \
                         calculate_new_variant_definition(left_read, right_read, fasta)
-                    if not low_confidence:
+                    if not failure_reason:
                         counter['Remapped'] += 1
                         info = left_read.query_name.split('|')
                         if info[7] != '.':
@@ -242,8 +242,8 @@ def process_bam_file(bam_file_path, output_file, out_failed_file, new_genome, fi
                     else:
                         # Currently the alignment is not precise enough to ensure that the allele change for INDEL and
                         # novel reference allele are correct. So we skip them.
-                        # TODO: add realignment confirmation
-                        counter['Low confidence event'] += 1
+                        # TODO: add realignment confirmation see #14 and EVA-2417
+                        counter[failure_reason] += 1
                         output_failed_alignment(primary_group, out_failed)
                 else:
                     output_failed_alignment(primary_group, out_failed)
--- a/variant_remapping_tools/reads_to_remapped_variants.py
+++ b/variant_remapping_tools/reads_to_remapped_variants.py
@@ -18,6 +18,9 @@ def calculate_new_variant_definition(left_read, right_read, ref_fasta):
     Resolve the variant definition from the flanking region alignment and old variant definition
     TODO: Link to algorithm description once public
     """
+    # Flag to highlight low confidence in an event detected
+    low_confidence = False
+
     # Grab information from the read name
     name = left_read.query_name
     info = name.split('|')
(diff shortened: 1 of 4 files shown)


tcezard:
Assessment results:

Type | TRUTH.TOTAL | TRUTH.TP | TRUTH.FN | QUERY.TOTAL | QUERY.FP | QUERY.UNK | FP.gt | METRIC.Recall | METRIC.Precision
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
INDEL | 383937 | 383907 | 30 | 383912 | 5 | 0 | 0 | 0.999922 | 0.999987
SNP | 3254376 | 3254260 | 116 | 3254263 | 3 | 0 | 0 | 0.999964 | 0.999999

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -220,16 +227,23 @@ def process_bam_file(bam_file_path, output_file, out_failed_file, new_genome, fi
...
+                            '%s\t%s\t%s\t%s\t%s\t%s\t%s\t%s\n' % (left_read.reference_name, varpos, info[4], new_ref,
+                                                                  ','.join(new_alts), info[5], info[6], info[7]))
                     else:
-                        info[7] = ';'.join(ops)
-                    outfile.write(
-                        '%s\t%s\t%s\t%s\t%s\t%s\t%s\t%s\n' % (left_read.reference_name, varpos, info[4], new_ref,
-                                                              ','.join(new_alts), info[5], info[6], info[7]))
+                        # Currently the alignment is not precise enough to ensure that the allele change for INDEL and
+                        # novel reference allele are correct. So we skip them.
+                        # TODO: add realignment confirmation
```
apriltuesday: what's realignment confirmation?

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -18,6 +18,9 @@ def calculate_new_variant_definition(left_read, right_read, ref_fasta):
     Resolve the variant definition from the flanking region alignment and old variant definition
     TODO: Link to algorithm description once public
     """
+    # Flag to highlight low confidence in an event detected
+    low_confidence = False
```
apriltuesday: Can this flag could be named something more specific?  I imagine we might want to keep track of the different reasons variants are discarded, but maybe that's not the intent of this flag.

apriltuesday (approved):
I'm kind of curious about the precision/recall changes from just the bug fix vs. just dropping low-confidence alignments if you have those numbers handy... Regardless, these are some impressive results!

variant_remapping_tools/reads_to_remapped_variants.py
tcezard: I'm not sure actually what it will be. 
I'm thinking some form of confirmation that the alleles (REF and ALTs) are the same before and after the remapping.
@tskir was suggesting Multiple Sequence Alignment.

variant_remapping_tools/reads_to_remapped_variants.py
tcezard: Do you mean having multiple flags ? for the "novel reference allele" and the "change in reference allele length"  ?
The main purpose of the flag is to convey that we have low confidence in those events. Not sure what else to name it.

apriltuesday: Yeah multiple flags for these might make more sense.  Basically I'm being influenced by the OT pipeline, where we track the different reasons records fail to produce evidence strings, some of which we can fix in the future and some we can't.  I could see something similar happening with variant remapping in the future as we try to improve it.

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -169,7 +174,9 @@ def pass_aligned_filtering(left_read, right_read, counter):
     :param counter: Counter to report the number of reads filtered.
     :return: True or False
     """
-    if left_read.cigartuples[-1][1] == 'S' or right_read.cigartuples[0][1] == 'S':
+    # in CIGAR tuples the operation is coded as an integer
+    # https://pysam.readthedocs.io/en/latest/api.html#pysam.AlignedSegment.cigartuples
+    if left_read.cigartuples[-1][0] == 4 or right_read.cigartuples[0][0] == 4:
```
```suggestion
    if left_read.cigartuples[-1][0] == pysam.CSOFT_CLIP or right_read.cigartuples[0][0] == pysam.CSOFT_CLIP:
```

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -220,16 +227,23 @@ def process_bam_file(bam_file_path, output_file, out_failed_file, new_genome, fi
             if pass_basic_filtering(primary_group, secondary_group, primary_to_supplementary, counter, filter_align_with_secondary):
                 left_read, right_read = _order_reads(primary_group, primary_to_supplementary)
                 if pass_aligned_filtering(left_read, right_read, counter):
-                    counter['Remapped'] += 1
-                    varpos, new_ref, new_alts, ops = calculate_new_variant_definition(left_read, right_read, fasta)
-                    info = left_read.query_name.split('|')
-                    if info[7] != '.':
-                        info[7] += ';'.join(ops)
+                    varpos, new_ref, new_alts, ops, low_confidence = calculate_new_variant_definition(left_read, right_read, fasta)
```
```suggestion
                    varpos, new_ref, new_alts, ops, low_confidence = \
                        calculate_new_variant_definition(left_read, right_read, fasta)
```

(Line over 120 columns)

variant_remapping_tools/tests/test_reads_to_remapped_variants.py
```diff
@@ -1,10 +1,11 @@
 import os
+from collections import Counter
 from unittest.mock import Mock, patch
 
 import pysam
 from unittest import TestCase
 from variant_remapping_tools.reads_to_remapped_variants import fetch_bases, process_bam_file, \
-    calculate_new_variant_definition, _order_reads, link_supplementary
+    calculate_new_variant_definition, _order_reads, link_supplementary, pass_aligned_filtering, group_reads
```
tskir: It was already like this before the change, but I'll point it out anyway: if `_order_reads` is being imported from that module, it shouldn't be a private method (starting with `_`)

variant_remapping_tools/reads_to_remapped_variants.py
tskir: Good point by @apriltuesday here, I agree having high granularity on error modes is very useful.

One possibility is to swap `low_confidence` for something like `reason_for_failure`, and to make it a _string_ rather than a bool. That string could then be fed directly into the counter in the calling function.

variant_remapping_tools/reads_to_remapped_variants.py
tskir: It actually took me ages to find this syntax because pysam documentation doesn't explicitly mention how this const is called...

FYI other attributes from the [list](https://pysam.readthedocs.io/en/latest/api.html#pysam.AlignedSegment.get_cigar_stats) are also available by stripping the `BAM_` prefix, e.g. `BAM_CEQUAL` becomes `pysam.CEQUAL`

variant_remapping_tools/reads_to_remapped_variants.py
tskir: To make it more explanatory perhaps we could link the relevant issue in the TODO: https://github.com/EBIvariation/variant-remapping/issues/24

tskir: I mean, the issue still just says “As described at the end of the Google document”, but I guess it's better than nothing :D

variant_remapping_tools/tests/test_reads_to_remapped_variants.py
```diff
@@ -59,19 +60,26 @@ def test_process_bam_file(self):
...
         ]
         with open(output_file, 'r') as vcf:
             for(i, line) in enumerate(vcf):
                 assert line == expected[i]
             # Expected and Generated VCF have the same number of lines
             assert i+1 == len(expected)
 
+    def test_pass_aligned_filtering(self):
+        left_read = self.mk_read(reference_start=0, reference_end=47, cigartuples=[(0, 47)])
+        right_read = self.mk_read(reference_start=51, reference_end=95,  cigartuples=[(4, 3), (0, 44)])
```
tskir: Suggest using the `pysam.CSOFT_CLIP`, `pysam.CEQUAL` syntax here also for clarity

tskir (commented):
The precision/recall metrics keep getting _even more amazing._ That's fantastic.

Did you have a chance to test on non-human data yet, BTW?

variant_remapping_tools/reads_to_remapped_variants.py
tcezard: Added reference to the issue and EVA ticket

variant_remapping_tools/reads_to_remapped_variants.py
tcezard: Ok that makes sense. I split the flag

variant_remapping_tools/tests/test_reads_to_remapped_variants.py
tcezard: Yes that didn't make sense I changed it

variant_remapping_tools/tests/test_reads_to_remapped_variants.py
tcezard: Add reference to the CONST. Thanks for pointing that out

variant_remapping_tools/reads_to_remapped_variants.py
apriltuesday: I've at least linked to the Google doc in question in the Github issue, for what it's worth :D

PR #27, opened 2021-05-07
EVA-1953 - Long read name

tcezard:
This PR fixes the issue with long vcf record (long INFO field or large INDELs) that cause the read name to become too long.

Instead of storing the VCF information in the read name I put the entire VCF record in the FASTA comment and replace the FASTA record name with a uniq id (line number). Using `-y` option in minimap2 I can transfer that comment to the output and use in the parsing.
I also took the opportunity to output all the fields from the original VCF including the genotypes.
I've done a quick test with 100K character line and it works.

It also add a simple method perform the genotype update for the currently supported usecase


README.md
```diff
@@ -1,13 +1,15 @@
 # variant-remapping
-Pipeline for remapping VCF variants between two arbitrary assemblies in FASTA format. No chain file is required.
+Pipeline for remapping VCF variants between two arbitrary assemblies in FASTA format. No chain file is required. 
+It does however assume that the source and destination genomes are closely related 
+and was designed with the explicit purpose of lifting over variants from one ersiong of the genome to another.
```
tskir:
```suggestion
Pipeline for remapping VCF variants between two arbitrary assemblies in FASTA format. No chain file is required. However, it does assume that the source and destination genomes are closely related and was designed with the explicit purpose of lifting over variants from one version of the genome to another.
```

README.md
```diff
@@ -46,30 +48,17 @@ nextflow run main.nf
     --newgenome <new_genome.fa> \
     --vcffile <source.vcf> \
     --outfile <remap.vcf> \
-    [--flankingseq 50] \
-    [--scorecutoff 0.6] \
-    [--diffcutoff 0.04]
```
tskir, line 51: After removing these lines, I think the final `\` on the previous line should be removed as well

main.nf
```diff
@@ -75,7 +75,7 @@ process storeVCFHeader {
         path "vcf_header.txt", emit: vcf_header
 
     """
-    bcftools view --header-only source.vcf | grep -v '^##FORMAT' >  vcf_header.txt
+    bcftools view --header-only source.vcf >  vcf_header.txt
```
```suggestion
    bcftools view --header-only source.vcf > vcf_header.txt
```

main.nf
```diff
@@ -168,7 +168,7 @@ process normalise {
         path "${outfile_basename}", emit: final_output_vcf
 
     """
-    bcftools norm -c ws -f genome.fa -N variants_remapped_sorted.vcf.gz -o ${outfile_basename} -O v
+    bcftools norm -f genome.fa  variants_remapped_sorted.vcf.gz -o ${outfile_basename} -O v
```
```suggestion
    bcftools norm --check-ref -f genome.fa  variants_remapped_sorted.vcf.gz -o ${outfile_basename} -O v
```

I'm not entirely sure about this suggestion, so this is more a point for discussion. My reasoning is:
* Because of the allele conversion algorithm, we expect that REFs will actually **always** align with the genome;
* If this is the case, it won't hurt to include `--check-ref` (equivalent to `-c e`; exit with an error if there is REF mismatch), as we expect it to do nothing;
* However, this can serve as an additional assert/safeguard that will fail loudly _in case_ something went wrong (e.g. due to some bug upstream)

tests/test_pipeline.sh
```diff
@@ -8,14 +8,16 @@ SOURCE_DIR=$(dirname $SCRIPT_DIR)
...
 cat << EOT > "${SCRIPT_DIR}/resources/source.vcf"
 ##fileformat=VCFv4.3
+##INFO=<ID=COMMENT,Number=1,Type=String,Description="Comment">
 ##FORMAT=<ID=GT,Number=1,Type=String,Description="Consensus Genotype across all datasets with called genotype">
-#CHROM	POS ID  REF ALT QUAL	FILTER	INFO	FORMAT	HG001
-chr1	48	.	C	A,T	50	PASS	.	GT	1/1
-chr1	98	.	C	CG	50	PASS	.	GT	1/1
+##FORMAT=<ID=GQ,Number=1,Type=Integer,Description="Genotype Quality">
+#CHROM	POS	ID	REF	 ALT	QUAL 	FILTER	INFO	FORMAT	HG001
+chr1	48	.	C	A,T	50	PASS	COMMENT=NANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANANABATMAN	GT:GQ	1/1:0
```
tskir: This is golden, I'm loving this :rofl:

variant_to_realignment.nf
```diff
@@ -100,42 +102,19 @@ process extractVariantInfoToFastaHeader {
...
-    # >[chr]|[pos interval]|[REF]|[ALT|[rsID]|[QUAL|[FILT]|[INFO]
-    # [seq]
-    sed 's/\\t/|/g; s/ /|/g; s/\\(.*\\)|/\\1 /' temp1.txt | tr ' ' '\\n' > variant_read1.out.fa
-    sed 's/\\t/|/g; s/ /|/g; s/\\(.*\\)|/\\1 /' temp2.txt | tr ' ' '\\n' > variant_read2.out.fa
+    # A space will be inserted between the position and the vcf fields
+    # Then a newline is inserted between the vcf fields and the sequence
+    # The vcf fields are regarded as comment to the fasta entry.
+    paste -d ' \\n' position.txt vcf_fields.txt <(grep -v '^>' variants_read1.fa)  > variant_read1.out.fa
+
+    paste -d ' \\n' position.txt vcf_fields.txt <(grep -v '^>' variants_read2.fa)  > variant_read2.out.fa
```
```suggestion
    paste -d ' \\n' position.txt vcf_fields.txt <(grep -v '^>' variants_read1.fa) > variant_read1.out.fa
    paste -d ' \\n' position.txt vcf_fields.txt <(grep -v '^>' variants_read2.fa) > variant_read2.out.fa
```

tskir (commented):
I like this! It's definitely another important step which brings the pipeline closer to production. 

Most comments are minor style nitpicks, with just one serious point for discussion (additional REF check in bcftools norm).

README.md
```diff
@@ -1,13 +1,15 @@
 # variant-remapping
-Pipeline for remapping VCF variants between two arbitrary assemblies in FASTA format. No chain file is required.
+Pipeline for remapping VCF variants between two arbitrary assemblies in FASTA format. No chain file is required. 
+It does however assume that the source and destination genomes are closely related 
+and was designed with the explicit purpose of lifting over variants from one ersiong of the genome to another.
 
 **Method**: creates reads from the flanking sequences of each variant, then maps them to the new assembly using 
-bowtie2.
+minimap2.
```
apriltuesday, line 5: Does this modification break the bowtie2 workflow?  (It's no problem to stop adding functionality or fixing bugs in the old workflow, but if we actively break it I would prefer removing the code completely or keeping it in a separate branch.)

main.nf
tcezard: I actually had the exact opposite thought: We know it'll always be the same as the reference so no need to check but I guess your point stands.

README.md
tcezard: It was, but bowtie2 has an option to support passing the fasta comment as well. 
I've added it so it should now work as well.
This said I'm now happy enough with minmap2 that I don't see the point of keeping bowtie2. 
I would drop it in a subsequent PR so it can be discuss there.

README.md
tcezard: Done

README.md
apriltuesday: makes sense, thank you!

apriltuesday (approved):

README.md
tskir: > This said I'm now happy enough with minmap2 that I don't see the point of keeping bowtie2.
I would drop it in a subsequent PR so it can be discuss there.

Sure, if you're happy with how minimap operates, then I don't see why we should bother to keep bowtie. So I support dropping it

variant_to_realignment.nf
```diff
@@ -17,10 +17,18 @@ process convertVCFToBed {
...
-    """
-    # TODO: Change vcf2bed so it does not split the alternates into two lines
-    vcf2bed < source.vcf > variants.bed
-    """
+    '''
+    # Convert the vcf file to bed format:
+    #  - Remove all headers
+    #  - Switch to 0 based coordinates system
+    #  - Add the reference allele so it can be used in flankingRegionBed to adjust the position of the right flank
+    #  - add all VCF fields separated by pipe (|)
```
tskir: On a second read, this is the only thing which still bothers me. A pipe symbol is commonly used in VCF to cram structured data into a string. For example, ClinVar:

```
1	925952	1019397	G	A	.	.	ALLELEID=1003021;CLNDISDB=MedGen:CN517202;CLNDN=not_provided;CLNHGVS=NC_000001.11:g.925952G>A;CLNREVSTAT=criteria_provided,_single_submitter;CLNSIG=Uncertain_significance;CLNVC=single_nucleotide_variant;CLNVCSO=SO:0001483;GENEINFO=SAMD11:148398;MC=SO:0001583|missense_variant;ORIGIN=1
```

I suspect that the `MC=SO:0001583|missense_variant` part will break this approach.

I don't see an easy solution to this. Using multiple pipes, e.g. `|||`, wouldn't be reliable either, because sometimes these structured values have quite a large number of fields, and some of them could be empty, and we can get something like `A|TRUE|0|||2||3.0`.

For now, as a quick fix, this could be perhaps addressed by using a multi-character separator which is unlikely appear in a real life VCF; so something mostly random.

And ultimately, I think it would be a good idea to have a deep refactor of this whole section of the code, so that the VCF records don't have to be stored in any other format (BED/BAM), and some different approach to pair the records is used instead. But this is clearly a lot of work and out of scope of this PR, and even perhaps the foreseeable future.

So, I leave the decision on how to handle this thing for you; other than that the changes look great :)

tcezard: That's a good point. It does not cost much more to use a two characters delimiter for now so I've implemented that.
I already implemented a [lookup system](https://github.com/EBIvariation/variant-remapping/blob/2abc2d0972a917fb603c5c01ef1cd74ac2850cce/variant_to_realignment.nf#L26) based on `seek` in this PR before, but it turned out to be quite slow.
Longer term I agree that a major refactor (or rewrite) is needed to get around this issue.

PR #28, opened 2021-05-24
EVA-2446 - Output unmapped variants

tcezard:
- Output the variant that could not be mapped in a separate file
- Add retry with more memory on kill for alignment stage

### DIFF
--- a/main.nf
+++ b/main.nf
@@ -39,6 +39,7 @@ if (!params.vcffile || !params.oldgenome || !params.outfile || !params.newgenome
 
 // basename and basedir of the output file to know how to name the output files
 outfile_basename = file(params.outfile).getName()
+outfile_basename_without_ext = file(params.outfile).getBaseName()
 outfile_dir = file(params.outfile).getParent()
 
 /*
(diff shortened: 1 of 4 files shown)


main.nf
```diff
@@ -86,14 +87,14 @@ include { process_split_reads; process_split_reads_mid; process_split_reads_long
 /*
  * Create the header for the output VCF
  */
-process buildHeader {
+process convertAndAddHeaderToVCF {
```
andresfsilva, line 90: what if we change the process name to something like `generateRemappedVCF` and add to the description that we are coverting to VCF and merging the header with the remapped variants?

main.nf
```diff
@@ -113,24 +114,29 @@ process buildHeader {
...
+    cat final_header.txt variants_remapped_sorted.vcf > variants_remapped_sorted_with_header.vcf
     """
 }
 
 /*
- * Add header to output VCF and merge with reference allele file
+ * Add header to unmapped variant VCF records
  */
-process mergeHeaderAndContent {
+process mergeOriginalHeaderAndVCFAndOutput {
```
andresfsilva, line 124: I think having the details in the comment description and a more general name helps to read the pipeline easier. Like renaming the process to something like `generateUnmappedVCF` and then explaining in the description that we are doing the merging

andresfsilva (approved):

apriltuesday (approved):

main.nf
tcezard: I've address this comment in subsequent PR [EVA-2451 - allow novel ref allele #29
](https://github.com/EBIvariation/variant-remapping/pull/29)

main.nf
tcezard: I've address this comment in subsequent PR [EVA-2451 - allow novel ref allele #29
](https://github.com/EBIvariation/variant-remapping/pull/29)

PR #29, opened 2021-05-26
EVA-2451 - allow novel ref allele

tcezard:
After testing the pipeline with real non-human data several issues crept up:
 - Too many variant were remapped to novel reference allele. So This PR re-enable them to be processed unless the reference allele length changes (shorter/longer indel or switch between variant types), also add genotypes change associated
- Some study name have space in them: replaced all spaces with double character "£€" while working wit the bed format. 
- Some variants are located too close to the edge of chromosomes preventing the construction of flanking regions: New filter function was added for the variants to be left aside. 
- Some remapping jobs have a lots of variants which is too slow to map: implemented a splitting of the reads in chunks to parallelise the process. To achieve this I had to put both flank of the variant in a single fasta file interleaved.


variant_to_realignment.nf
```diff
@@ -56,7 +57,7 @@ process flankingRegionBed {
         | bedtools slop  -g genome.chrom.sizes -l $flankingseq -r 0  > flanking_r1.bed
 
     # Adjust the start position of the flank to be one base downstream of the end of variant (\$4 is the reference allele)
```
```suggestion
    # Adjust the start position of the flank to be one base downstream of the end of variant (\$5 is the reference allele)
```

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -98,7 +99,9 @@ def update_vcf_record(reference_name, varpos, new_ref, new_alts, operations, ori
             genotype_str_list = original_vcf_rec[genotype_i].split(':')
             if genotype_str_list[gt_index] == '1/1':
                 genotype_str_list[gt_index] = '0/0'
-                original_vcf_rec[genotype_i] = ':'.join(genotype_str_list)
+            elif 'nra' in operations and genotype_str_list[gt_index] == '0/1':
+                genotype_str_list[gt_index] = '1/2'
+            original_vcf_rec[genotype_i] = ':'.join(genotype_str_list)
```
apriltuesday, line 105: I know it's probably a lot of work, but I think tests for this method (in particular the genotype changes) would be quite valuable.

variant_to_realignment.nf
tcezard: Fixed

variant_remapping_tools/reads_to_remapped_variants.py
tcezard: I've added some minimal test.

apriltuesday: awesome, thanks!

apriltuesday (approved):

tests/test_pipeline.sh
```diff
@@ -2,6 +2,15 @@
 
 set -Eeuo pipefail
 
+function asserteq() {
```
apriltuesday, line 5: we're going to end up writing our own testing framework for nextflow in bash, aren't we...

main.nf
```diff
@@ -235,24 +279,49 @@ workflow finalise {
...
+            params.newgenome,
+            prepare_new_genome_bowtie.out.genome_fai,
+            prepare_new_genome_bowtie.out.bowtie_indexes
+        )
+        finalise(process_split_reads_with_bowtie.out.variants_remapped, storeVCFHeader.out.vcf_header,
+                 params.newgenome, process_split_reads_with_bowtie.out.summary_yml)
+}
+
+
+
```
apriltuesday, line 311:
```suggestion
```

apriltuesday (approved):

PR #31, opened 2021-06-16
Allow the read to remap step to be run in parallel and only merge after

tcezard:
When using scatter gather, only gather after the `readsToRemappedVariants` step

### DIFF
--- a/variant_remapping_tools/reads_to_remapped_variants.py
+++ b/variant_remapping_tools/reads_to_remapped_variants.py
@@ -1,6 +1,5 @@
 #! /usr/bin/env python3
 import argparse
-import time
 from argparse import RawTextHelpFormatter
 from collections import Counter, defaultdict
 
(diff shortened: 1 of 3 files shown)


sundarvenkata-EBI (approved):

variant_to_realignment.nf
```diff
@@ -270,9 +269,32 @@ process readsToRemappedVariants {
                 -o variants_remapped.vcf  --newgenome genome.fa --out_failed_file variants_unmapped.vcf \
                 --flank_length $flank_length --summary summary.yml
            """
+}
+
+/*
+ * Take the reads and process them to get the remapped variants
+ *
```
apriltuesday, line 276: comment should be updated

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -1,5 +1,6 @@
 #! /usr/bin/env python3
 import argparse
+import time
```
```suggestion
```

apriltuesday (approved):

PR #32, opened 2021-08-20
Remove variant that have non ATGC bases in the reference

tcezard:
I've encountered a case of a variant that remapped perfectly on a non ATCG base which fails at the normalisation stage. 
It does not make sense to remap a C->G variant onto an S (which mean C or G) so I'd rather fail these case.

### DIFF
--- a/variant_remapping_tools/reads_to_remapped_variants.py
+++ b/variant_remapping_tools/reads_to_remapped_variants.py
@@ -29,7 +29,7 @@ def calculate_new_variant_definition(left_read, right_read, ref_fasta, original_
                           right_read.reference_start - left_read.reference_end).upper()
 
     if len(set(new_ref).difference(nucleotide_alphabet)) != 0 :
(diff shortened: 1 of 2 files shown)


sundarvenkata-EBI (approved):

variant_remapping_tools/reads_to_remapped_variants.py
```diff
@@ -27,6 +28,9 @@ def calculate_new_variant_definition(left_read, right_read, ref_fasta, original_
     new_ref = fetch_bases(ref_fasta, left_read.reference_name, left_read.reference_end + 1,
                           right_read.reference_start - left_read.reference_end).upper()
 
+    if len(set(new_ref).difference(nucleotide_alphabet)) != 0 :
+        failure_reason = 'Reference Allele not in non ACGT'
```
```suggestion
        failure_reason = 'Reference Allele not in ACGT'
```

apriltuesday (approved):

PR #35, opened 2022-07-13
EVA-2919 : Filter out Novel Reference Allele

nitin-ebi:


main.nf
```diff
@@ -186,21 +186,37 @@ process sortVCF {
...
+    output:
+        path "variants_remapped_sorted_normalized.vcf.gz", emit: variants_remapped_sorted_normalized_gz
+
+    """
+    bcftools norm --check-ref e -f genome.fa --old-rec-tag PRE_NORM variants_remapped_sorted.vcf.gz -o variants_remapped_sorted_normalized.vcf.gz -O z
+    """
+}
+
+
+process nraAndOutput {
```
sundarvenkata-EBI:
```suggestion
process collectNovelReferenceAlleles {
```

sundarvenkata-EBI (approved):

tcezard (commented):
This is all correct but I thought that we would be adding that outside of the variant-remapping. Into the [variant remapping automations](https://github.com/EBIvariation/eva-tools/tree/master/variant-remapping-automation).

main.nf
nitin-ebi: done

nitin-ebi:
> This is all correct but I thought that we would be adding that outside of the variant-remapping. Into the [variant remapping automations](https://github.com/EBIvariation/eva-tools/tree/master/variant-remapping-automation).

apologies, when you said "we need to do it outside of variant-remapping", I thought you meant outside of the code that does the actual remapping. But as discussed in the stand-up, leaving it as it is.

PR #40, opened 2024-06-05
EVA-3330: Add labels to nextflow for SLURM migration

apriltuesday:


variant_to_realignment.nf
```diff
@@ -150,13 +150,11 @@ process split_fasta {
...
 
-    // Memory required is 5 times the size of the fasta in Bytes or at least 1GB
-    // Retry on kill (exit status 130) with twice the amount of memory
-    memory { Math.max(file(params.newgenome).size() * 10, 2000000000) * task.attempt + ' B' }
-
-    errorStrategy { task.exitStatus == 130 ? 'retry' : 'terminate' }
-    maxRetries 3
+    // Memory required is 10 times the size of the fasta in Bytes or at least 2GB
+    // Overwrite base_memory so that the standard retry strategy is used
+    ext base_memory: { Math.max(file(params.newgenome).size() * 10, 2000000000) }
```
apriltuesday, line 157: This is now working with updated configs, see [run on Seqera](https://cloud.seqera.io/orgs/EMBL-EBI/workspaces/ebi-spws-dev-eva-submission/watch/10mFkwcTSHB0tS)

tcezard: This is good but I'm wondering how anyone else that does not have our nextflow config would use this.
We could add some default nextflow config to the repo or add to the documentation how the memory requirement should be managed.

tcezard (approved):

variant_to_realignment.nf
apriltuesday: Yes this is a good point, a config placed alongside the pipelines would (I think) take precedence over the one in the home directory which I'm not sure is a good idea for us. But including an example config somewhere and documenting the usage is definitely a good idea.

apriltuesday: Added, [here](https://github.com/EBIvariation/variant-remapping/pull/40/commits/fe250f7f4266b6c89e33ecaf97371733562088d1)

nitin-ebi (approved):

tcezard:
Looks good to me

